PR #113, opened 2026-05-11
fix(console): preserve credentials for local OSS API requests

mindburnlabs:
Resolves local authentication issues for the OSS console by properly forwarding credentials when hitting a localhost or user-configured endpoint.

This PR strictly scopes changes to the OSS boundary and explicitly removes any dependencies on commercial DO proxies or hosted infrastructure.


eipp:
Closed as superseded by #114 and the prod follow-up #115. The console credential-preservation fix is already on main, and this older branch now carries stale workflow changes that should not be merged into production.

PR #115, opened 2026-05-12
Prod follow-up: unblock main workflow health

peycheff-com:
Fixes post-merge workflow health after the forced deployment merge.\n\n- Remove top-level env from Scorecard workflow so published results verify on main pushes.\n- Move hash-locked codegen requirements out of a directory that Dependabot graph treats as an unsupported pip manifest directory.\n- Update CI/release references and docs coverage accordingly.\n\nLocal checks in clean worktree:\n- actionlint .github/workflows/ci.yml .github/workflows/release.yml .github/workflows/scorecard.yml\n- make docs-coverage (passed); docs-truth only failed because the temporary worktree basename differs from repo manifest identity\n- pip --dry-run --require-hashes for .github/codegen-requirements.txt\n- git diff --check


eipp (approved):
Approving production workflow hotfix after local workflow and requirements validation.

PR #119, opened 2026-05-12
feat(design-system): add public site styles

mindburnlabs:
## Summary
- add shared public-site navigation and execution-demo CSS primitives
- expose HELM token compatibility aliases for downstream marketing surfaces
- normalize touch targets for status metadata

## Tests
- make test-design-system
- git diff --check


eipp (approved):
Approved for final branch cleanup after local design-system verification.

PR #120, opened 2026-05-12
fix(connectors): close TON Acton gaps

mindburnlabs:
## Summary
- update TON Acton command mappings and supported current Acton/Tolk versions
- add receipt decision hashes and schema coverage
- add proofgraph/tape evidence entries and env-gated real Acton smoke test

## Tests
- go test ./core/pkg/connectors/ton/acton ./core/pkg/contracts ./core/pkg/conformance
- git diff --check HEAD


eipp (approved):
Approved after local TON Acton connector verification.

PR #121, opened 2026-05-12
feat(console): version agent ui route

mindburnlabs:
## Summary
- add /api/v1/agent-ui routes while keeping /api/ag-ui compatibility routes
- update console runtime defaults and generated API schema
- add shared button, icon, skip-link, frame, and footer design-system styles
- refresh boundary manifest

## Tests
- go test ./core/cmd/helm
- make test-design-system
- make test-console
- make verify-boundary
- git diff --check


eipp (approved):
Approved after local console, design-system, boundary, and Go verification.

PR #128, opened 2026-05-12
ci: fetch main for tag release proto checks

mindburnlabs:
## Summary\n- Fetches a local main ref during tag-triggered Release validation.\n- Fixes buf breaking in detached tag checkouts where .git#branch=main was unavailable.\n\n## Validation\n- make proto-breaking\n- Release failure confirmed on v0.5.0 was due to missing local main ref in tag checkout.


eipp (approved):
Reviewed release workflow tag-checkout fix.

PR #132, opened 2026-05-12
ci: fix release image and benchmark pinning

mindburnlabs:
## Summary\n- use lowercase GHCR image path for release container publishing\n- write generated benchmark reports to the root benchmarks/results path expected by release pinning\n- ignore generated latest benchmark report while keeping per-release snapshots pinnable\n\n## Validation\n- make bench-report\n- bash scripts/release/pin_benchmarks.sh v0.5.0


eipp (approved):
Approved final release workflow fixes after local benchmark report/pin validation.

PR #135, opened 2026-05-13
ci: use multiarch alpine base for slim image

mindburnlabs:
## Summary
- Replace the slim image Alpine platform-specific digest with the multi-arch manifest digest for alpine:3.21.
- Fixes the v0.5.0 release slim image failure where arm64 used an amd64-only base and failed under QEMU with Invalid ELF image.

## Validation
- docker buildx imagetools inspect alpine:3.21
- docker buildx build --platform linux/amd64,linux/arm64 -f Dockerfile.slim --output type=cacheonly .


eipp (approved):
Approved. This directly fixes the observed slim image arm64 base mismatch; local two-platform Buildx validation passed.

PR #136, opened 2026-05-13
ci: authenticate cosign container signing

mindburnlabs:
## Summary
- Log in to GHCR before keyless cosign signing of release container images.
- Fixes the v0.5.0 release failure where cosign could obtain OIDC credentials but could not pull the private/protected GHCR image reference for signing.

## Validation
- Previous release run confirmed main and slim images build/push successfully.
- Failure log showed GHCR token request returned UNAUTHORIZED during cosign sign.
- This patch supplies the workflow GITHUB_TOKEN to docker/login-action before cosign.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -232,6 +232,11 @@ jobs:
       packages: write
       id-token: write
     steps:
+      - uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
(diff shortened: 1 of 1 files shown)


eipp (approved):
Approved. This directly addresses the release cosign-container failure: cosign reached Fulcio/Rekor flow but GHCR pull was unauthorized, so the job needs a registry login before signing.

PR #145, opened 2026-05-17
feat(chart): publish to ghcr.io as OCI artifact, fix tag format

SergeyAP:
Adds a chart job to release.yml that packages deploy/helm-chart and pushes
it to oci://ghcr.io/mindburn-labs/charts on every tag, using GITHUB_TOKEN
(write:packages is already granted). The chart job runs after container
and cosign-container, and github-release now depends on chart, so a failed
chart push blocks the release.

Bumps chart version to 0.5.1 and aligns appVersion to "v0.5.0" so the
rendered Deployment matches the actual image tag in ghcr.io. release.yml
publishes containers as github.ref_name, which always carries the leading
"v"; previously a default `helm install` would have failed with
ImagePullBackOff against a non-existent :0.5.0 image.

This enables the ArtifactHub repository at mindburn-labs to discover the
chart automatically once the next tag is pushed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -310,8 +310,8 @@ jobs:
       contents: read
       packages: write
     steps:
-      - uses: actions/checkout@v4
-      - uses: azure/setup-helm@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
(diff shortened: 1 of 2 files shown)


peycheff-com (approved):
Approved for branch consolidation. Checks are green; the remaining blocker was the required review gate.

PR #155, opened 2026-05-18
ci: accept encoded Maven GPG keys

peycheff-com:
## Summary
- Make Maven publish GPG import accept raw armor, base64-encoded key material, or escaped-newline armor
- Keep the publish workflow compatible with the existing maven-central environment secret

## Validation
- workflow YAML parse
- git diff --check


mindburnlabs (approved):
Approving Maven GPG import hardening for production publish.

eipp (approved):
Second approval after last push for Maven GPG import compatibility.

PR #157, opened 2026-05-18
chore(release): prepare v0.5.2 with refreshed ArtifactHub registration

SergeyAP:
Targeted release-prep on top of the v0.5.1 baseline now that PR #149 (`d19a107`) shipped the canonical export/verify fix via dynamic `copyIndexedEvidenceEntries()`. This branch only carries the bits that were **not** in #149:

- Bump `VERSION` and `Chart.yaml::version`/`appVersion` to `0.5.2` so the kernel binary, chart, and tag all line up.
- Refresh `artifacthub-repo.yml` `repositoryID` to `5cafa88d-…` — the new ArtifactHub registration created after the previous OCI repo URL was corrected from `oci://ghcr.io/mindburn-labs/charts` to `oci://ghcr.io/mindburn-labs/charts/helm-ai-kernel`. AH now indexes the chart under repo `mindburn-labs`; this UID is what lights up Verified Publisher once we expose it as an OCI artifact.
- Add the ArtifactHub repository badge to `README.md`.
- Drop the `>/dev/null` redirect from export and verify in `stage_release_assets.sh`. `d19a107` added structured `log_step` output but kept the stdout suppression, so the verify failure that crashed the v0.5.1 binaries job would still be invisible if it recurred.

---

> **Note (added 2026-05-19):** the branch `release/v0.5.2-rebuild` now also contains commit `74f9867` (`fix(kernel,chart): close boundary fail-open, …`), which was pushed on top of this PR by a separate session. That commit is being moved to its own PR #158. Once #158 lands first, this PR's diff will be reduced to just the release-prep commit above.

PR #158, opened 2026-05-18
fix(kernel,chart): close boundary fail-open, anchor KMS on dataDir, wire build-info ldflags, retire phantom metrics port

SergeyAP:
## Summary

Four-part fix-stack uncovered during the post-`v0.5.1` smoke pass of the chart from `artifacthub.io`. The same commit also rides on `release/v0.5.2-rebuild` as part of PR #157, but it is presented here on its own branch (off `origin/main`, single commit) so it can be reviewed and merged independently of the v0.5.2 release-prep work.

- **Boundary fail-open (P0)** — `services.go` hardcoded `Version: "1.0"` while `boundary.PolicyVersion` is `"1.0.0"`. `LoadPolicy` does strict equality, so the init returned an error, the kernel logged it as `WARN`, the very next log line falsely claimed the subsystem was `initialized`, and `BoundaryEnforcer` stayed `nil`. With a nil enforcer, every `CheckNetwork` / `CheckTool` / `CheckData` silently passes — a fail-closed firewall that was actually fail-open. Switched to the canonical constant via a new `defaultBoundaryPolicy()` helper, and hardened `main.go` so the kernel refuses to start in `HELM_PRODUCTION=1` when the default policy fails to load (instead of swallowing the error as "degraded mode").
- **KMS keystore on read-only rootfs (P1)** — `NewLocalKMS` was opened against the relative path `"data/keys/credentials.keystore.json"`, which on distroless resolves under the read-only rootfs (CWD=`/`), not under the PVC-mounted dataDir. The init silently disabled the credentials store. Anchored on the same `dataDir` that drives sqlite/signer via a new `kmsKeystorePath()` helper (explicit arg → `HELM_DATA_DIR` env → `"data"`). The chart now also exports `HELM_DATA_DIR` alongside `--data-dir` so any future env-only subsystem stays consistent.
- **Build-info ldflags (P3)** — All four Dockerfiles (`Dockerfile`, `Dockerfile.slim`, `core/Dockerfile`, `core/Dockerfile.api`) shipped with `-ldflags="-s -w"` and never injected `main.version` / `main.commit` / `main.buildTime`, so `GET /version` returned `"unknown"` for everything but the Go runtime. Added `BUILD_VERSION` / `BUILD_COMMIT` / `BUILD_TIME` ARGs and propagated them via `-X`. `release.yml` computes the values in a new `build-meta` step (strips the leading `v` so `main.version` stays semver) and passes them to both `docker/build-push-action` invocations.
- **Phantom metrics port (P2)** — `values.yaml` advertised port `9090` and a `ServiceMonitor` scraping it, but the kernel never registered `metrics.Handler()`. Port-forward to 9090 was always refused. Disabled metrics in defaults (templates were already correctly gated) with a `NOTE` pointing at the follow-up task `chart-metrics-backend-implementation`. The phantom port comes back when the backend lands.

## Test plan

- [x] `go test ./...` — full kernel suite green, including two new regression tests:
  - `core/cmd/helm-ai-kernel/boundary_default_test.go` — locks in the boundary schema version contract.
  - `core/cmd/helm-ai-kernel/kms_keystore_path_test.go` — covers all three resolution branches.
- [x] `helm lint deploy/helm-chart` — clean.
- [x] `helm template deploy/helm-chart` — Service exposes only `http` + `health`; no `metrics` port by default.
- [x] Built `helm-ai-kernel:dev-smoke` from `core/Dockerfile` with `BUILD_VERSION=0.5.2-dev BUILD_COMMIT=346aa5836cb1 BUILD_TIME=2026-05-18T22:19:55Z` and ran the chart in minikube with `helm.production=false`:
  - Pod boots without `KMS init failed — credentials store DISABLED` and without `Boundary enforcer init error="unsupported version: 1.0"`. Subsystems log straight to `Credentials Handler initialized (KMS-backed)` and `Boundary Perimeter Enforcer initialized`.
  - `GET /version` → `{"version":"v0.5.2-dev","commit":"346aa5836cb1","build_time":"2026-05-18T22:19:55Z","go_version":"go1.25.10"}` (real values, not `"unknown"`).
  - `GET /api/v1/boundary/check` → `{"enforcer":"active","status":"ready"}` (was `{"status":"disabled"}` on v0.5.1).
  - `POST /v1/chat/completions` with a real `OPENAI_TOKEN` and `model: gpt-4o-mini` → `403 Governance Blocked / PDP_DENY (ref=helm:helm-chart)`. The token was **not** spent — no upstream call in pod logs. Receipt persisted via `GET /api/v1/receipts` with the existing `lamport_clock` / `prev_hash` chain and ed25519 signature.
- [x] CI workflows (`ci.yml`, `release.yml`) green on this PR.

## Notes for review

- The new fatal-on-production behaviour is gated strictly on `HELM_PRODUCTION=1`. The chart default `helm.production: false` keeps minikube / dev smoke working as before — only production deployments refuse to start when boundary init fails.
- `NewServices` signature now takes `dataDir string` as the final argument. Only one caller in this repo (`main.go::runServerWithOptions`); external embedders, if any, will see a breaking change.
- The phantom-metrics retirement is a values default change, not a hard removal. Operators with `--set helm.metrics.enabled=true` will still get the (unbound) port until the backend lands — kept that way intentionally so the `chart-metrics-backend-implementation` follow-up can flip the default back without further template changes.

## Relationship to PR #157

PR #157 (branch `release/v0.5.2-rebuild`) currently contains both this commit and the release-prep commit `346aa58`. Once this PR merges into `main`, PR #157's diff will reduce to just the release-prep commit — at that point #157 can be merged or this PR can be considered the v0.5.2 contents and #157 closed as merged-into-main.

Tag `v0.5.2` (annotated, already pushed) points at the combined snapshot on `release/v0.5.2-rebuild` and is kept intentionally — the release.yml workflow has already started building artifacts from that tag, and the snapshot is what v0.5.2 will release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -210,6 +210,17 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
+      - id: build-meta
+        name: Compute build metadata
+        run: |
+          # Strip the leading "v" from a tag ref so main.version stays semver.
+          ref="${GITHUB_REF_NAME}"
+          if [[ "${ref}" =~ ^v[0-9] ]]; then ref="${ref#v}"; fi
+          {
+            echo "version=${ref}"
+            echo "commit=$(git rev-parse --short=12 HEAD)"
+            echo "build_time=$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+          } >> "${GITHUB_OUTPUT}"
       - id: push
         name: Build and push main image
         uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6
@@ -220,6 +231,10 @@ jobs:
           provenance: true
           sbom: true
           platforms: linux/amd64,linux/arm64
+          build-args: |
+            BUILD_VERSION=${{ steps.build-meta.outputs.version }}
+            BUILD_COMMIT=${{ steps.build-meta.outputs.commit }}
+            BUILD_TIME=${{ steps.build-meta.outputs.build_time }}
           tags: |
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.ref_name }}
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
@@ -233,6 +248,10 @@ jobs:
           provenance: true
           sbom: true
           platforms: linux/amd64,linux/arm64
+          build-args: |
+            BUILD_VERSION=${{ steps.build-meta.outputs.version }}
+            BUILD_COMMIT=${{ steps.build-meta.outputs.commit }}
+            BUILD_TIME=${{ steps.build-meta.outputs.build_time }}
           tags: |
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.ref_name }}-slim
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,12 @@
 FROM --platform=$BUILDPLATFORM golang:1.25.10-alpine@sha256:8d22e29d960bc50cd025d93d5b7c7d220b1ee9aa7a239b3c8f55a57e987e8d45 AS builder
 ARG TARGETOS
 ARG TARGETARCH
+# Build metadata injected via ldflags into main.{version,commit,buildTime} and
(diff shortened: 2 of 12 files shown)


mindburnlabs (approved):
Release-admin review: audited the security-critical boundary/KMS/chart/release metadata patch, verified PR checks are green, and ran focused local validation: go test ./cmd/helm-ai-kernel -run 'TestDefaultBoundaryPolicyLoads|TestKMSKeystorePath', go test ./..., and make helm-chart-smoke.

eipp (approved):
Second release-admin/code-owner approval after independent audit of PR #158 and successful CI/local validation. Approving for protected-branch merge.

PR #164, opened 2026-05-19
ci: extend Maven Central publish wait

peycheff-com:
## Summary
- extend Sonatype Central publish polling from 30 minutes to 2 hours
- reduce polling noise by checking every 15 seconds

## Verification
- mvn -q test package
- git diff --check

## Context
The corrected Maven deployment uploaded successfully under `io.github.mindburnlabs`, then the Central plugin failed because the default 30-minute `waitMaxTime` elapsed before Sonatype reported the deployment as published.

### DIFF
--- a/sdk/java/pom.xml
+++ b/sdk/java/pom.xml
@@ -112,6 +112,8 @@
                     <publishingServerId>central</publishingServerId>
                     <autoPublish>true</autoPublish>
                     <waitUntil>published</waitUntil>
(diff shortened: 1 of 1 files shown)


mindburnlabs (approved):
Approved. This addresses the Central publish timeout after successful bundle upload.

eipp (approved):
Approved. Longer wait window matches Central's observed publication latency.

PR #207, opened 2026-05-26
fix(launchpad,ci): restore launchpad foundation after main rebase

SergeyAP:
## Summary

Force-push to main on 2026-05-26 dropped several launchpad hardening fixes that had landed in #204 / #206 and pre-existing `gap-*` notes in the daily log. As a result every `oss_supported` local-container app fails at runtime start, while `docs/launchpad/CONFORMANCE.md` continues to assert the opposite. This PR re-applies the minimum set needed to make the foundation match the canon — with one opportunistic lint-gate fix so the PR can be green.

### Foundation fixes

- **`.github/workflows/ci.yml`** — guard `git fetch origin main:refs/heads/main` against push-to-main (the `contract-drift` job's git fetch is the regression we previously fixed in #206).
- **`core/pkg/launchpad/session/runtime_starter.go` + `healthcheck_runner.go`** — `egressProxyFromEnv` now takes the substrate ID and refuses the `LaunchOwnedEgressProxy` default (host loopback, unreachable from container netns) for substrates that isolate the network namespace. Replaces the previous silent `OpenRouter key check failed with HTTP status 000` masquerade with an actionable error.
- **`core/pkg/sandbox/docker/runner.go`** — mount tmpfs at `/tmp` alongside `--read-only`, matching the existing `sandboxes.go` path. Without this every workload writing to `/tmp` (openclaw, hermes, Python configs) fails with EROFS before doing useful work.
- **`core/pkg/conform/evidencepack.go`** — promote `11_HOST_EVIDENCE` to a §3.1 canonical subdirectory. The launchpad writer has been emitting this section since `ac243ad`; the conformer's whitelist was the drifted half. `helm-ai-kernel verify --bundle` (the very command the kernel suggests in every LaunchRun) was returning FAILED on every launchpad pack.

### Yaml fixes for headless launch

- **`registry/launchpad/apps/openclaw.yaml`** — bare `[openclaw]` drops into the Crestodian TUI and exits 1 without a terminal. Replaced with `[openclaw, gateway, run, --port, 18789, --allow-unconfigured, --auth, none, --bind, loopback]`, the foreground HTTP gateway form.
- **`registry/launchpad/apps/hermes.yaml`** — bare `[hermes]` hits a broken venv shebang (`#!/src/...` not `#!/opt/...` in the published image) and exits with `exec: no such file or directory`. Wrapped in `/bin/sh -c` with `HOME=/tmp/hermes-home` to keep the writable-home redirect on tmpfs.

### Coverage manifest

- **`docs/documentation-coverage.csv`** — add a row for the `observability/` surface that was added in `cf16fd9` without a coverage row, currently blocking the `Coverage and truth` and `kernel` jobs.

### Formatting

- **`core/cmd/helm-ai-kernel/launch_cmd.go`** — gofmt-only import sort to satisfy `make lint`.

## What this does NOT yet fix

Validated live on Intel mac (Intel-amd64, colima) with all changes applied:

- **openclaw** gateway now actually starts (`[gateway] ready` after ~30s), but the kernel runtime timeout (120s hardcoded in `local_container.go`) still trips because the combined warmup + healthcheck exceeds it, and plugins fail because `/opt/openclaw/.openclaw` is on read-only rootfs without an `app_state` mount. Tracked as gap #12 (configurable timeout) + gap #19 (`app_state:rw` runtime enforcement).
- **hermes** upstream CLI in the current pinned image rejects `-z 'healthcheck ping'` with `Could not consume arg: -z`. The `runtime.command` wrapper survives but the healthcheck needs a CLI surface the new image accepts. Tracked separately.

Both of the above need further investigation; the user explicitly asked to ship this foundation PR first and continue debugging them in a follow-up.

## Test plan

- [x] `make test` (full kernel suite) — green locally
- [x] `make lint` — green locally
- [x] 4 new unit tests for `egressProxyFromEnv` (static URL, sidecar image, fail-fast refusal, no-allowlist returns nil)
- [x] Live-verified on Intel mac (`sergey@192.168.1.203`, Intel-amd64 colima):
  - `helm-ai-kernel verify --bundle <pack>.tar` now returns **VERIFIED** on launchpad-generated packs (was: `FAILED · unexpected top-level entry: 11_HOST_EVIDENCE`)
  - openclaw with `[openclaw, gateway, run, ...]` reaches `[gateway] ready` in ~30s (was: `Crestodian needs an interactive TTY` exit 1)
  - hermes with `/bin/sh -c` wrapper progresses past exec to the actual CLI (was: `exec /opt/hermes/.venv/bin/hermes: no such file or directory` exit 255)
- [ ] CI on this PR goes green
- [ ] CI on `main` after merge — `contract-drift` job no longer fails on push-to-main (the actual gate this PR's CI fix addresses)


SergeyAP:
Superseded by #208

PR #243, opened 2026-06-02
Cover pack and trust follow-up edges

mindburnlabs:
Adds follow-up coverage for pack registry/readability edge cases, verifier signature status handling, and trust pack loader validation branches.

Verification:
- go test ./pkg/pack ./pkg/trust -count=1
- git diff --check


mindburnlabs:
Addressed Codex review: the pack loader TUF fixture now signs each metadata payload with the trusted Ed25519 private key instead of using a dummy signature. Verified locally with `go test ./pkg/pack ./pkg/trust -count=1` from `core/` and `git diff --check`.

PR #275, opened 2026-06-08
feat(helm-chart,launchpad): co-deploy openclaw and hermes Pods in k8s with smoke CI

SergeyAP:
## Summary

Adds a Kubernetes co-deployment path that mirrors the local-container runbook
for openclaw and hermes onto a chart-managed Pod topology, without introducing
a launchpad `kubernetes` substrate (that scope remains on
`launchpad-k8s-substrate-kernel-governance` umbrella, milestones M1-M4).

- openclaw and hermes pinned via sha256 digest, share the helm release with
  the kernel, front their outbound traffic through a per-Pod egress sidecar
  bound to localhost.
- `launchpadApps.{openclaw,hermes}` value blocks gated by `enabled` (default
  `false` — baseline chart-install behaviour preserved).
- Hermes runs as a short-lived Job; openclaw as a Deployment with the
  `helm-launchpad-openrouter-check` readiness probe (same script used by
  local-container substrate via docker exec).
- NetworkPolicy default-deny + allow only sidecar → external; no app-to-app
  in-cluster traffic.
- `helm.sh/hook: test` template (`launchpad-connectivity.yaml`) checks
  app reachability post-install.

## Out of scope

- No new launchpad substrate (`kubernetes`). Apps are deployed via chart,
  kernel is not in the loop for runtime governance in this PR.
- No multi-arch (amd64 only via `nodeSelector: kubernetes.io/arch=amd64`).
- No differential check `compose ≡ k8s`.
- No MCP runtime quarantine — covered by separate
  `openclaw-extended-functional-coverage` pt. 2.

## Test plan

- [x] Cherry-picked clean onto current `origin/main` HEAD `3523461` from
  prior `feat/launchpad-k8s-smoke` branch (4 commits, no manual conflicts)
- [x] Live: validated on intel-mac stand via \`minikube\`-driven
  \`scripts/ci/launchpad_k8s_smoke.sh\` (full pass including readiness
  probes); see commit \`8c92ef2 fix(helm-chart,launchpad): land live-verify
  findings\`
- [ ] CI: \`helm-integration.yml\` workflow runs \`{baseline, positive, negative}\`
  matrix on this PR
- [ ] Reviewer: confirm helm chart Lint/Schema clean, NetworkPolicy default-deny
  matches expected semantics


SergeyAP:
## Egress enforcement — addressing the CAP_NET_ADMIN review

**What changed (commit `68f3940`).** The launchpad egress sidecar was honor-based: the workload only got `HTTP_PROXY`, so a direct connection to `:443` left the pod, bypassing the sidecar and leaving no receipt. This replaces it with a **transparent proxy**:

- an `egress-init` container (`CAP_NET_ADMIN`) installs an iptables `REDIRECT` that funnels all outbound TCP from the workload into the sidecar;
- the sidecar recovers the original destination via `SO_ORIGINAL_DST` (hostname via TLS SNI, best-effort), enforces the allowlist, and writes a receipt for **every** attempt — allow and deny;
- the sidecar runs under a dedicated uid (65532), exempted from the REDIRECT to avoid a loop; `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` are dropped from the app containers;
- the egress-proxy image gains `iptables` (one image, two roles — init + sidecar).

**Why.** Under a prompt-injection threat model, honor-based egress is not a control — a compromised agent simply ignores the proxy env. Transparent redirect makes bypass structurally impossible and, unlike a plain iptables `REJECT`, routes even *denied* attempts through the receipt writer, so the evidence pack records them.

**Verified on minikube/calico (real OpenRouter key):**
- direct `:443` from the openclaw container (1.1.1.1, api.github.com, registry.npmjs.org) → **blocked** (TLS reset) + a DENY receipt each; pre-fix these returned 200/301 with no receipt;
- `openrouter.ai` → reachable transparently (ALLOW + receipt), no proxy env;
- hermes Job completes; `egress-init` exits 0 (no nft conflict on calico).
- Reports: [transparent-enforce verify](https://github.com/Mindburn-Labs/docs_for_team/blob/main/REPORTS/2026-06-10_pr275_egress-transparent-enforce_verify/LOG.md) · [smoke](https://github.com/Mindburn-Labs/docs_for_team/blob/main/REPORTS/2026-06-10_pr275_smoke/REPORT.md).

**Scope & caveats.**
- SNI is best-effort (TLS 1.3 ECH / IP-literal / non-TLS → receipt keyed by IP). Only **TCP** egress is intercepted; DNS/UDP/ICMP are out of scope (separate control). Docs say "all **TCP** egress", not "all egress".
- Follow-up: the baked openclaw readiness script still requires `HTTP_PROXY` and fails the probe — it must drop that guard before openclaw reaches Ready (hermes is unaffected).
- The `egress-proxy` digest in `values.yaml` is still the old CONNECT build; it will be bumped after CI rebuilds the image.

**CI pinning (Scorecard / Pinned-Dependencies, code-scanning 229/230/231).** Fixed in `6c3af71` — the `get-helm-3 | bash` and `minikube …/latest` download-then-run are replaced with version-pinned downloads (minikube v1.38.1, helm v3.21.0) verified by sha256 before install; versions/hashes live in workflow `env`.

PR #292, opened 2026-06-10
fix(release): publish homebrew formula to canonical tap + daily drift watch

eipp:
**Root cause of the stale homebrew tap (v0.5.6 while kernel released v0.5.10):** the release workflow's `homebrew` job cloned/pushed `mindburnlabs/homebrew-tap` — the wrong org slug — and the published-drift contract (`release/version-surfaces.yaml`) watched the same wrong URL. The fan-out and its watchdog were both aimed at a repo that doesn't exist under that name, so releases never updated the real tap and the daily `version-drift.yml` watcher couldn't see the drift.

**Fix (2 files):** correct the org slug to `Mindburn-Labs/homebrew-tap` in `release.yml` (clone + push) and in the `homebrew-tap` surface of `version-surfaces.yaml`. The existing daily `version-drift.yml` reads this contract, so alerting is repaired by the same change — an initially-proposed second watcher workflow was dropped as duplicate truth.

Companion: Mindburn-Labs/homebrew-tap#1 manually brings the tap to 0.5.10 (awaiting review approval).

Verification: `python3 scripts/release/check_version_drift.py --expected-version 0.5.10 published` output attached in comments after tap#1 merges; workflow YAML parse + quality gates in CI.


peycheff-com (approved):
Reviewed: scoped, matches audit FINAL_REPORT slice; CI green so far. Approving per owner authorization.

PR #326, opened 2026-06-11
fix(helm-chart): re-pin egress-proxy digest to a build with iptables

SergeyAP:
## Problem

PR #275 introduced the `egress-init` transparent-proxy step (runs `iptables ... REDIRECT`) and added `RUN apk add --no-cache iptables` to `egress-proxy.Dockerfile`. But the **same commit** (`be62955e`) pinned `launchpadApps.{hermes,openclaw}.egressSidecar.image.digest` to `sha256:e09e0aec…` — a build made *before* that Dockerfile change.

The deployed `egress-init` therefore fails immediately:

```
/bin/sh: iptables: not found     # exit 127
```

It never installs the redirect rules → pod goes `Init:Error` → **every launchpad app with an egress sidecar (hermes, openclaw) fails to start.**

## Fix

Re-pin both `egressSidecar` digests to `sha256:c55a277a35…` (tag `enforce`), the current published egress-proxy image that **does** include iptables. No rebuild needed — the correct image is already in GHCR.

## Found by

functional-test run `2026-06-12_ftmain-bcef_smoke-llm-containerd` (containerd telemetry validation). Re-run case 03 (agent-bootstrap) after merge to confirm `egress-init` passes and hermes reaches the model provider.

## Follow-up

Digest-drift prevention (auto-pin launchpad image digests from CI) tracked separately.

### DIFF
--- a/deploy/helm-chart/values.yaml
+++ b/deploy/helm-chart/values.yaml
@@ -337,7 +337,7 @@ launchpadApps:
     egressSidecar:
       image:
         repository: "ghcr.io/mindburn-labs/helm-launchpad/egress-proxy"
-        digest: "sha256:e09e0aec1e0e1f926f4cd18444e88310656b85551cbc10a6c340acb979a42e03"
+        digest: "sha256:c55a277a350ee715cb1cbfb017aa20025df5b4f717b283e0144b1c3bf0887537"
         pullPolicy: "IfNotPresent"
       port: 8080
       allowlist:
@@ -378,7 +378,7 @@ launchpadApps:
     egressSidecar:
       image:
         repository: "ghcr.io/mindburn-labs/helm-launchpad/egress-proxy"
-        digest: "sha256:e09e0aec1e0e1f926f4cd18444e88310656b85551cbc10a6c340acb979a42e03"
+        digest: "sha256:c55a277a350ee715cb1cbfb017aa20025df5b4f717b283e0144b1c3bf0887537"
         pullPolicy: "IfNotPresent"
       port: 8080
       allowlist:


SergeyAP:
## Verification — functional-test on minikube / containerd

Ran this fix on minikube (containerd). Two things:

⚠️ **The first committed digest `c55a277` is not pullable.** It was a stand-local
egress-proxy image that was never pushed to GHCR — the node pulls **anonymously by
digest** and gets `NotFound` → `ImagePullBackOff` on every launchpad egress sidecar.

✅ **Corrected to `sha256:22d78073…`** (commit `4d007de`) — the egress-proxy image that
this PR's parent merge (#275) *actually* built and published (tag `be62955e0692…` = the
merge commit SHA; built 2026-06-10; `/usr/sbin/iptables` present; anonymously pullable).
Net diff of this PR vs `main` is now `e09e0aec → 22d78073`.

Verified in-cluster with the corrected digest:
- `egress-init` → `Completed (0)` for **hermes and openclaw** (was `exit 127: iptables: not found`).
- Case 03 (agent-bootstrap) graceful degrade: Node.js install blocked by the egress allowlist
  (`curl (35) nodejs.org:443`), agent still answered via `gemini-3.5-flash`.

Root cause is digest drift: #275 added `apk add iptables` to the Dockerfile but pinned the
**pre-merge** `e09e` in values.yaml, while the merge build published the correct `22d78073`.
Follow-up to auto-pin launchpad digests from CI is tracked separately.

Full report (lands shortly):
https://github.com/Mindburn-Labs/docs_for_team/blob/main/REPORTS/2026-06-12_pr326_egress-digest-fix-verify/REPORT.md

PR #329, opened 2026-06-12
fix(launchpad): route openclaw agent model through OpenRouter

SergeyAP:
## Problem

The **`Helm Integration — Launchpad k8s smoke`** workflow has been red on `main` for days. The failing job is **`openclaw + hermes positive (real OPENROUTER_API_KEY)`** → step *Run positive smoke* (exit 1). `chart-only baseline` and `openclaw + hermes negative (fake key)` are green; hermes itself works (it runs `--provider openrouter`).

This is **not** a regression from a single PR — it is a long-standing blocker that surfaced after the egress-init `iptables` fixes (#326/#327) let the run progress far enough to reach the model call.

## Root cause

openclaw's gateway starts `--allow-unconfigured` and falls back to its built-in default **agent model `openai/gpt-5.5`**, which resolves to the **`openai/*` provider and dials `api.openai.com` directly**:

```
[gateway] agent model: openai/gpt-5.5 (thinking=medium, fast=off)
[agent/embedded] embedded run failover ... from=openai/gpt-5.5
  rawError=stream disconnected before completion:
  error sending request for url (https://api.openai.com/v1/responses)
Embedded agent failed before reply: stream disconnected before completion ...
```

But openclaw's egress allowlist is **openrouter-only** (`policies/launchpad/apps/openclaw.safe.toml` → `network_allowlist = ["https://openrouter.ai/api/v1"]`), enforced by the transparent-egress iptables REDIRECT. So the call to `api.openai.com` is dropped and the embedded agent fails. The launchpad wires up OpenRouter (key, allowlist, readiness probe) but never tells the openclaw **agent** to route through it.

## Fix

Point openclaw's OpenAI-compatible client at OpenRouter (which is OpenAI-API-compatible) and authenticate with the same OpenRouter key, in `openclaw-deployment.yaml`:

```yaml
- name: OPENAI_BASE_URL
  value: "https://openrouter.ai/api/v1"
- name: OPENAI_API_KEY
  valueFrom:
    secretKeyRef:
      name: {{ .Values.openrouter.apiKeySecretRef.name }}
      key: {{ .Values.openrouter.apiKeySecretRef.key }}
```

Now openclaw's agent calls leave through the only allowlisted host. This mirrors **hermes**, which already runs `--provider openrouter` and passes.

## Open item / validation

This redirect is the necessary routing fix. One thing to confirm on CI: openclaw's default model slug `openai/gpt-5.5` must resolve on OpenRouter. If OpenRouter returns `400 model not found`, a follow-up must pin a known slug (e.g. `openai/gpt-4o-mini`, as hermes uses) — but openclaw's model-selection knob is not exposed in this repo (external image, `--allow-unconfigured`), so that would need an openclaw config/env from the upstream image.

**Validation:** the `Helm Integration — Launchpad k8s smoke` workflow (positive job) is the authoritative check. Also run locally on the intel stand via `/functional-test helm-ai-kernel minikube` before merge.

### DIFF
--- a/deploy/helm-chart/templates/launchpad-apps/openclaw-deployment.yaml
+++ b/deploy/helm-chart/templates/launchpad-apps/openclaw-deployment.yaml
@@ -80,6 +80,22 @@ spec:
                 secretKeyRef:
                   name: {{ .Values.openrouter.apiKeySecretRef.name | quote }}
                   key: {{ .Values.openrouter.apiKeySecretRef.key | quote }}
+            # Route openclaw's OpenAI-compatible agent calls through OpenRouter —
+            # the only host the egress allowlist permits. openclaw's built-in agent
+            # model defaults to the `openai/*` provider, which dials api.openai.com
+            # directly; under the transparent-egress REDIRECT that connection is
+            # dropped ("stream disconnected before completion"), so the embedded
+            # agent fails. OpenRouter is OpenAI-API-compatible, so pointing
+            # OPENAI_BASE_URL at it (authenticated with the same OpenRouter key)
+            # makes those calls leave through the allowed host. Mirrors hermes,
+            # which already runs `--provider openrouter`.
+            - name: OPENAI_BASE_URL
+              value: "https://openrouter.ai/api/v1"
+            - name: OPENAI_API_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.openrouter.apiKeySecretRef.name | quote }}
+                  key: {{ .Values.openrouter.apiKeySecretRef.key | quote }}
           ports:
             - name: gateway
               containerPort: {{ $cfg.gatewayPort }}


SergeyAP:
## Stand validation (minikube launchpad, openclaw positive) — this fix does **not** work ❌

Validated PR #329 on minikube (reproducing the red CI job `openclaw + hermes positive`). Three findings:

**1. The `OPENAI_BASE_URL` approach is ineffective.** With `OPENAI_BASE_URL=https://openrouter.ai/api/v1` and `OPENAI_API_KEY` confirmed present in the running openclaw container, the agent run still dialed `https://api.openai.com/v1/responses` (dropped by the openrouter-only transparent egress → `stream disconnected before completion`). openclaw does **not** consult `OPENAI_BASE_URL` for the `openai/*` provider, and it uses the OpenAI **Responses API** (`/v1/responses`), which OpenRouter doesn't serve. → not mergeable as-is.

**2. The real reason the CI positive smoke is red is a *separate* stale-digest bug.** openclaw never becomes Ready: the readiness probe `helm-launchpad-openrouter-check` exits 43 *"Launchpad egress proxy missing"* even though `HELM_EGRESS_TRANSPARENT=1` is set. The probe **baked into the pinned openclaw image** (`@sha256:4da80a1e…`) is the pre-#328 version (checks only `HTTP(S)_PROXY`, no `HELM_EGRESS_TRANSPARENT` branch). #328 fixed the probe *source* but the openclaw image was never rebuilt + re-pinned in `values.yaml` — digest-drift. This is the `assert_pod_ready` failure; it needs an openclaw image re-pin (and a look at why `pin-digests` didn't fire for openclaw after #328).

**3. The correct routing fix — validated.** openclaw routes via its **native config**, not env:
```
openclaw config set agents.defaults.model openrouter/openai/gpt-4o-mini
```
openclaw then auto-enables the OpenRouter provider from the model prefix + `OPENROUTER_API_KEY` (already in the deployment). After this, an agent turn returned **`pong`** via OpenRouter with no `api.openai.com`. The deployable form is to inject this into the openclaw container (mount an `openclaw.json`, or prepend `openclaw config set …` to the command before `gateway run`).

**Recommendation:** convert this PR to **Draft**; reland the routing fix as the `agents.defaults.model` config change (replacing `OPENAI_BASE_URL`), **paired with the openclaw image digest re-pin** — both are needed to green the positive smoke.

Full report (lands on push to `docs_for_team` `main`):
https://github.com/Mindburn-Labs/docs_for_team/blob/main/REPORTS/015_2026-06-12_pr329_minikube_openclaw-openrouter/REPORT.md

eipp:
Adversarial review (production-readiness closeout) — **holding, not merging on the green rollup**:

- The authoritative job `openclaw + hermes positive (real OPENROUTER_API_KEY)` was **skipped** on this PR head: its `if:` requires a non-PR event or the `launchpad-live-test` label, and the label set was empty — so the only job that exercises this change never ran. I've now added the `launchpad-live-test` label and updated the branch (was behind_by=32) so the positive job runs against current main.
- The green negative test exercises the egress *readiness* probe, a path separate from the openclaw agent's `openai/*` model call — it does not validate the routing fix.
- The egress-allowlist correctness checks out (openrouter.ai:443 ALLOW, api.openai.com:443 DENY), but the `openai/gpt-5.5` model-resolution risk is unsettleable from the diff.

@SergeyAP — please confirm the live-test job goes green with the real `OPENROUTER_API_KEY` secret; if it returns `400 model-not-found`, a model-pin follow-up is needed before merge. Holding for your sign-off.

PR #332, opened 2026-06-12
feat(launchpad): publish multi-arch app manifests

peycheff-com:
## Summary
- Publish launchpad app and egress-proxy artifacts as `linux/amd64,linux/arm64` OCI manifest candidates.
- Keep the chart default `launchpadApps.nodeSelector.kubernetes.io/arch: amd64` because the chart-pinned openclaw/hermes/egress-proxy digests are still amd64-only until a follow-up digest re-pin lands.
- Let `scripts/ci/launchpad_k8s_smoke.sh` use `KUBE_HELM_CMD` when `helm` is occupied by the HELM verifier, and make the launchpad image preload path explicit debug-only by platform.
- Keep the positive OpenRouter CI job visible as a required check, but do not burn tokens on ordinary PRs unless the `launchpad-live-test` label is present and the secret is configured.

## Validation
- `bash -n scripts/ci/launchpad_k8s_smoke.sh scripts/ci/helm_chart_smoke.sh`
- `/usr/bin/python3 -m json.tool deploy/helm-chart/values.schema.json`
- `actionlint .github/workflows/helm-integration.yml .github/workflows/launchpad-artifacts.yml`
- `PATH="/tmp/kube-helm.6nvp4p/bin:$PATH" bash scripts/ci/helm_chart_smoke.sh`
- `helm template` assertion: all 4 rendered pod specs carry `imagePullSecrets[0].name=ghcr-read`; all 3 launchpad pod specs keep `nodeSelector.kubernetes.io/arch=amd64`.

## MIN-494
This does not complete MIN-494. The issue should remain `Merged/Verifying` until an actual Hermes/OpenRouter co-deploy smoke runs with a real key and records nonzero tokens via the docs_for_team token-meter path.

### DIFF
--- a/.github/workflows/helm-integration.yml
+++ b/.github/workflows/helm-integration.yml
@@ -81,10 +81,11 @@ jobs:
 
   launchpad-positive:
     name: openclaw + hermes positive (real OPENROUTER_API_KEY)
-    if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'launchpad-live-test') }}
     runs-on: ubuntu-latest
     timeout-minutes: 45
     needs: [baseline]
+    env:
+      RUN_LIVE_POSITIVE: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'launchpad-live-test') }}
     steps:
       - name: Checkout
         uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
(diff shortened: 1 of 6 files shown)


peycheff-com:
Added follow-up commit `61eb9017` intentionally keeps `launchpadApps.nodeSelector.kubernetes.io/arch: amd64` until the chart-pinned launchpad digests are re-pinned to multi-arch indexes.

Please do not update/rewrite the branch back to the single-commit version that leaves `nodeSelector: {}` while the current pinned digests are still amd64-only. That would make Apple Silicon/arm64 scheduling fail in the gap before the re-pin PR lands.

The positive smoke check is also kept visible but non-burning on ordinary PRs; MIN-494 still stays `Merged/Verifying` until a real Hermes/OpenRouter co-deploy smoke burns measurable tokens.

PR #361, opened 2026-06-13
ci: require OpenRouter token proof for live smoke

mindburnlabs:
## Summary
- add a stdlib OpenRouter activity delta meter for CI release evidence
- wrap the Helm Integration live positive smoke with token measurement
- fail main/dispatch live positives when the OpenRouter secret is missing or no token delta is observed
- upload redacted token evidence artifacts for release-truth reports

## Validation
- python3 -m unittest scripts/ci/test_openrouter_token_meter.py
- python3 scripts/ci/openrouter_token_meter.py --help
- parsed .github/workflows/helm-integration.yml with PyYAML
- git diff --check

## Release-truth note
This is required before REPORTS/014 can be a GO RC proof: the live OpenRouter smoke must have token accounting instead of only a successful job conclusion.

PR #363, opened 2026-06-13
fix(ci): separate OpenRouter proof from smoke command

peycheff-com:
## Summary
- Fix the Helm Integration positive smoke shell block so the OpenRouter token probe and launchpad smoke script run as separate commands.
- Preserve #361's token-proof gate while unblocking main smoke.

## Validation
- /usr/bin/ruby -e 'require "yaml"; YAML.load_file(".github/workflows/helm-integration.yml")'
- /usr/bin/python3 -m unittest scripts/ci/test_openrouter_token_meter.py
- command-separation assertion for --min-total and launchpad_k8s_smoke.sh
- git diff --check


peycheff-com:
Superseded by merged PR #362, which applies the same Helm Integration command separation fix. Closing this duplicate branch and continuing validation on current main.

PR #449, opened 2026-06-30
Add privacy-preserving risk envelope contract

mindburnlabs:
## Summary
- harden `core/pkg/riskenvelope` with CSPRNG local salt files, canonical `envelope_content_hash`, suppression metadata, and Go/schema enum parity tests
- add `core/pkg/riskscan` to project raw `shadow.Report` + local Claude/Codex/MCP config observations into an upload-safe RiskEnvelope, Markdown/HTML previews, deterministic evidence-pack tar, and explicit upload helper
- add top-level `helm-ai-kernel scan` with `--risk-envelope`, repeatable `--preview`, `--evidence-pack`, and guarded `--upload --upload-url ... --yes`
- add `helm-ai-kernel scan --from-receipts <dir>` to project workstation observe/decision receipts into the same anonymized RiskEnvelope without touching runtime dispatch
- document the operator path in `docs/reference/agent-risk-scan.md`, README, developer journey, CLI reference, schema reference, and docs coverage ledger
- move the old top-level `scan` alias out of `shadow`; `helm-ai-kernel shadow scan` remains the raw local scanner

## Privacy boundary
- local scan and receipt projection export RiskEnvelope-derived fields only
- no raw paths, repo names, prompt text, command bodies, source snippets, metadata, targets, secret values, or local salts are serialized into previews, evidence packs, or uploads
- upload remains explicit `--upload --upload-url ... --yes`, and prints body hash/size/destination before sending

## Validation
- `env -u GOROOT -u GOBIN /usr/local/go/bin/go test ./pkg/riskenvelope ./pkg/riskscan ./pkg/shadow ./cmd/helm-ai-kernel`
- `env -u GOROOT -u GOBIN PATH="/usr/local/go/bin:$PATH" make verify-boundary`
- `PATH="/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/opt/homebrew/bin:/usr/local/go/bin" make docs-coverage docs-truth`
- `git diff --check`

## Notes
This is the local-first product wedge plus observe-mode bridge. It does not add a backend ingest route or enforce-mode runtime dispatch behavior.

PR #452, opened 2026-06-30
Publish risk scan docs route

mindburnlabs:
## Summary
- add the new Agent Risk Scan reference page to the public docs manifest
- preserves source-owned docs publication authority for helm.docs.mindburn.org generation

## Validation
- make verify-boundary
- git diff --check

## Context
Follow-up to #449 so app-helm-docs can publish /reference/agent-risk-scan from the merged source doc.

### DIFF
--- a/docs/public-docs.manifest.json
+++ b/docs/public-docs.manifest.json
@@ -433,6 +433,15 @@
       "sensitivity": "public"
     },
     {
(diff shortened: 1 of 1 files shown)


peycheff-com (approved):
Approved manifest-only docs publication fix. Local validation passed: make verify-boundary, make docs-coverage docs-truth, and git diff --check.

PR #458, opened 2026-07-01
docs: keep publication policy machine output public-safe

peycheff-com:
## Summary
- remove exact restricted route strings from the public publication-policy source page
- keep the operational requirement that reserved/private namespaces and unknown routes fail closed with noindex
- prevents llms-full.txt from leaking customer/internal/Enterprise route terms after docs regeneration

## Validation
- /usr/bin/python3 scripts/check_documentation_truth.py
- make verify-boundary
- git diff --check


eipp (approved):
Reviewed the machine-output boundary hotfix after CI rerun. It removes restricted public/private route strings from publication-policy machine output without weakening the live route deny policy.

mindburnlabs (approved):
Second approval for the public-docs machine-output boundary hotfix. CI is green after retry and the change is limited to removing restricted route vocabulary from public artifacts.

PR #473, opened 2026-07-02
Pass only docs truth read token

peycheff-com:
## Summary
- repins the docs-truth caller to the explicit-secret reusable workflow
- passes only MINDBURN_ORG_READ_TOKEN instead of inheriting every repo secret

## Validation
- actionlint .github/workflows/docs-truth.yml
- git diff --check


mindburnlabs (approved):
Approved after caller workflow now pins the reusable workflow SHA and passes only MINDBURN_ORG_READ_TOKEN.

PR #476, opened 2026-07-02
fix(release): allow Python publish from release tag

mindburnlabs:
## Summary
- add an optional `release_tag` input to the manual Python Publish workflow
- validate that `release_tag` is exactly `v<version>` before any tag checkout
- allow current workflow fixes to publish a prior tagged Python SDK after `main` moves forward

## Why
The v0.5.20 PyPI recovery currently depends on `main` still carrying `VERSION=0.5.20`. Once `main` is prepared for v0.6.0, the current manual workflow could no longer publish `helm-sdk==0.5.20` with the restored token path. This keeps recovery possible without using the old tag workflow.

Example recovery dispatch after the PyPI account/token blocker is fixed:

```bash
gh workflow run python-publish.yml \
  --repo Mindburn-Labs/helm-ai-kernel \
  --ref main \
  -f version=0.5.20 \
  -f release_tag=v0.5.20 \
  -f dry_run=false
```

## Validation
- `MISE_TRUSTED_CONFIG_PATHS=/Users/ivan/Code/Mindburn-Labs/helm-ai-kernel python3 scripts/ci/validate_workflow_dispatch_inputs_test.py`
- `actionlint -shellcheck= .github/workflows/python-publish.yml`
- `git diff --check`

## Release truth
This does not complete v0.5.20 or start v0.6.0. It only keeps PyPI recovery possible after v0.6.0 prep begins.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -12,6 +12,11 @@ on:
         required: false
         default: false
         type: boolean
+      release_tag:
+        description: Optional v<version> tag to publish with the current workflow.
+        required: false
+        default: ""
+        type: string
 
 permissions: read-all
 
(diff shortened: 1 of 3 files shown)


eipp (approved):
Approved: narrow workflow recovery path. The release_tag input is constrained to v<version>, and validator/actionlint checks passed locally.

peycheff-com (approved):
Approved current head fb9b662 after local validator/actionlint/diff checks and all required GitHub checks passed. Scope is limited to preserving PyPI recovery for a prior release tag; no release claim.

PR #479, opened 2026-07-02
Expose receipt signature posture metadata

mindburnlabs:
## Summary
- add receipt signature posture metadata fields to kernel receipts and OpenAPI/SDK types
- stamp Ed25519, hybrid Ed25519+ML-DSA-65, and ML-DSA-only receipt signatures with profile, algorithm, key id, and public key set
- make profile-aware receipt verification reject declared profile/algorithm mismatches and verify ML-DSA-only pqc receipts
- update MCP bundle/benchmark wording away from Ed25519-only receipt claims
- refresh the protected boundary manifest for the touched receipt/crypto files

## Validation
- git diff --check
- go test ./core/pkg/contracts ./core/pkg/crypto
- make quantum-crypto-inventory
- make verify-boundary
- make sdk-openapi-check
- make quality-pr
- GOCACHE=/Users/ivan/Code/Mindburn-Labs/helm-ai-kernel/.cache/go-build make test-sdk-go-standalone (rerun outside sandbox because httptest needs localhost bind)

## Release/status note
- This is source/default-branch candidate work only; it does not claim v0.6.0 is released or published.

PR #482, opened 2026-07-02
Publish proof loop docs in manifest

mindburnlabs:
## Summary
- adds public docs manifest entries for the now-tracked proof loop and AI security categories pages
- keeps the docs-source gate aligned with the files restored in #479

## Validation
- git diff --check
- python3 -m json.tool docs/public-docs.manifest.json


peycheff-com (approved):
Approved manifest-only follow-up after local docs validation: git diff --check, json.tool, and make docs-coverage docs-truth.

PR #486, opened 2026-07-02
docs: use neutral tenant KMS heading

peycheff-com:
## Summary
- Replace public-docs source heading/text that would be rewritten from Enterprise to advanced
- Keep tenant KMS posture wording neutral and accurate: downgrade-aware classical Ed25519, not PQ key material

## Validation
- python3 scripts/check_documentation_truth.py
- python3 scripts/ci/check_quantum_crypto_inventory.py --base origin/main
- git diff --check


mindburnlabs (approved):
Approved after all required kernel/docs/security checks passed; this follow-up fixes the public docs heading so generated docs do not rewrite it to misleading wording.

PR #490, opened 2026-07-02
Fix Homebrew release publishing through tap PR

mindburnlabs:
## Summary

- Change the release workflow Homebrew job from direct push to protected `homebrew-tap` main to a release-specific branch + PR flow.
- Make the job idempotent when the tap formula already matches the release tag.
- Attempt the protected-branch merge through the tap PR path so tag releases can complete the Homebrew gate.

## Validation

- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file("/Users/ivan/Code/Mindburn-Labs/helm-ai-kernel-wt-release-v0.6.0/.github/workflows/release.yml")'`
- `actionlint .github/workflows/release.yml` was run; it still reports pre-existing shellcheck findings in unrelated blocks, not in the modified Homebrew block.

## Release context

This fixes the failure observed on the `v0.5.20` release rerun where the Homebrew job could create the formula commit but was blocked by tap rules requiring changes through a pull request.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -334,7 +334,7 @@ jobs:
           mkdir -p tap/Formula
           cp Formula/helm-ai-kernel.rb tap/Formula/helm-ai-kernel.rb
           cd tap
-          if git diff --quiet -- Formula/helm-ai-kernel.rb; then
+          if [ -z "$(git status --porcelain -- Formula/helm-ai-kernel.rb)" ]; then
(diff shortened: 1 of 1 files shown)


peycheff-com (dismissed):
Approved: release workflow now uses the Homebrew tap PR path instead of direct-pushing to protected main; scoped to the v0.5.20 Homebrew failure mode.

peycheff-com (approved):
Approved after follow-up: the no-op guard now uses git status --porcelain, so untracked first-time formula files are not silently skipped.

PR #494, opened 2026-07-03
docs: make policy bundle signing profile-aware

peycheff-com:
## Summary
- Remove Ed25519-only policy-bundle signing language from the RFC.
- Add explicit signature_profile/signature_algorithm fields to the manifest example.
- Document classical vs hybrid bundle-signature profiles and downgrade rejection for hybrid deployments.

## Validation
- rg stale Ed25519-only policy-bundle strings: no matches
- git diff --check
- make quantum-crypto-inventory
- make docs-truth


mindburnlabs (approved):
Reviewed current head afff6230. Scope is docs plus regenerated protected boundary manifest; local make quality-pr, docs-truth, quantum crypto inventory, and diff check pass.

PR #495, opened 2026-07-03
Document console hybrid receipt display posture

mindburnlabs:
## Summary
- Updates the Kernel-owned quantum posture doc with the console hybrid receipt display proof boundary.
- Keeps authenticated live workspace receipt smoke listed as a separate remaining gate.

## Validation
- make docs-truth
- make docs-coverage
- make quantum-crypto-inventory
- git diff --check

### DIFF
--- a/docs/security/quantum-posture.md
+++ b/docs/security/quantum-posture.md
@@ -1,6 +1,6 @@
 ---
 title: Quantum Posture
-last_reviewed: 2026-07-02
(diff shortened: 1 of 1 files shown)


peycheff-com (approved):
Approved after the console receipt wording was corrected to distinguish raw Kernel contracts.Receipt fields from the control-plane/console projection.

PR #496, opened 2026-07-03
fix(guardian): preserve snapshot policy hash under embedded PDP (HELM-49)

Hirama:
## Problem
When an `EffectivePolicySnapshot` carries an embedded PDP, `PDPInterceptor` leaves `PDPHash` empty (the snapshot owns the binding). The final handler then overwrote the snapshot-bound `PolicyContentHash` with that empty value, so every ALLOW under snapshot+PDP failed `IssueExecutionIntent` with "decision missing policy hash/epoch binding".

## Fix
Only overwrite `PolicyContentHash` from the PDP when the PDP actually reported a hash. Regression test covers snapshot+embedded-PDP through intent issuance.

Fixes HELM-49.

### DIFF
--- a/core/pkg/guardian/guardian.go
+++ b/core/pkg/guardian/guardian.go
@@ -956,7 +956,12 @@ func (g *Guardian) EvaluateDecision(ctx context.Context, req DecisionRequest) (*
 
 		if eCtx.PDPBackend != "" {
 			decision.PolicyBackend = eCtx.PDPBackend
-			decision.PolicyContentHash = eCtx.PDPHash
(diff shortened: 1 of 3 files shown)


mindburnlabs (approved):
Approved current rebuilt head after conflict resolution. Local validation on current origin/main passed: go test ./core/pkg/guardian, make verify-boundary, git diff --check origin/main...HEAD, and make quality-pr. Scope is limited to preserving the snapshot-bound policy hash when an embedded PDP reports no hash, plus regression coverage and a narrow protected-manifest refresh; this is source/kernel validation only, not release or runtime evidence.

PR #498, opened 2026-07-03
fix(guardian): require trusted context for tainted-egress override (HELM-52)

Hirama:
## Problem
`allow_tainted_egress` was honored directly from caller-supplied request context, letting any agent self-approve egress of PII/credential/secret tainted data.

## Fix
The override now applies only when the context carries the `security_context_trusted` binding set by a trusted transport/adapter boundary — the same gate as `credential_hash` and `session_id`. Updates the coverage test that asserted the insecure behavior; adds end-to-end tests for both ignored-untrusted and honored-trusted overrides.

Note: the whole gate remains opt-in via `HELM_TAINT_TRACKING`; flipping that default is left as a policy decision in HELM-52.

Fixes HELM-52 (caller-bypass half).

### DIFF
--- a/core/pkg/guardian/guardian.go
+++ b/core/pkg/guardian/guardian.go
@@ -1085,7 +1085,11 @@ func taintedEgressDenied(ctx map[string]interface{}, labels []string) bool {
 	if len(labels) == 0 || ctx == nil {
 		return false
 	}
-	if approved, ok := ctx["allow_tainted_egress"].(bool); ok && approved {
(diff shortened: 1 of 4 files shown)


mindburnlabs (approved):
Approved current rebuilt head 538aafe5 after local validation: go test ./core/pkg/guardian, make verify-boundary, git diff --check origin/main...HEAD, and make quality-pr passed. Scope is limited to requiring trusted security context for allow_tainted_egress plus focused regression coverage and boundary manifest refresh.

PR #499, opened 2026-07-03
fix(guardian): burn delegation nonce only after scope checks (HELM-53)

Hirama:
## Problem
`MarkNonceUsed` ran before the tool-scope and per-action checks. A legitimate delegated request that hit a scope violation (e.g. a tool typo) consumed the single-use nonce, so the corrected retry failed as a false replay — DoS on valid delegated sessions.

## Fix
Move the nonce burn to the success path, after every delegation check passes. Regression test: a scope-denied attempt leaves the nonce intact and the corrected retry is allowed.

Fixes HELM-53.

### DIFF
--- a/core/pkg/guardian/guardian_delegation_test.go
+++ b/core/pkg/guardian/guardian_delegation_test.go
@@ -149,6 +149,51 @@ func TestGuardian_Delegation_ToolScopeViolation_Deny(t *testing.T) {
 	assert.Contains(t, decision.Reason, "forbidden_tool")
 }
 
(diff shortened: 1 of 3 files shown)


mindburnlabs (approved):
Approved current rebuilt head dc197bdd96818c8e07701037d213502302c19709 after local validation: go test ./core/pkg/guardian, make verify-boundary, git diff --check origin/main...HEAD, and make quality-pr. Scope verified as nonce burn order plus focused regression and touched-file boundary manifest refresh only; no release/runtime claim.

PR #500, opened 2026-07-03
fix(kernel): guard InMemoryEffectBoundary with a mutex (HELM-54)

Hirama:
## Problem
The reference effect boundary held its maps with no synchronization. Concurrent `Submit` calls raced the maps (Go fatal "concurrent map writes") and the check-then-set idempotency logic could admit two effects for one key, double-executing the side effect.

## Fix
Add a `sync.Mutex` covering every map access; make the idempotency check-and-reserve atomic within `Submit`. Adds a `-race` concurrency test asserting a single effect per shared key.

Fixes HELM-54.


eipp (approved):
Approved current head after branch update. Local validation on a clean rebuild from current origin/main passed: go test -race ./core/pkg/kernel, make verify-boundary, git diff --check, and make quality-pr. This is source/kernel correctness only, not release or runtime evidence.

mindburnlabs (approved):
Approved current head as requested reviewer after all required checks passed. Reviewed scope is limited to the kernel InMemoryEffectBoundary mutex/idempotency fix plus regenerated protected manifest; this is source/kernel validation only, not release or runtime evidence.

PR #502, opened 2026-07-03
fix(guardian): consume budget after policy, fail closed (HELM-50)

Hirama:
## Problem
- Budget was consumed during the pre-policy check, so a PRG-denied request had already drawn down budget.
- A `Consume` error was logged and swallowed, allowing the effect on an unrecorded consumption.

## Fix
Keep the fast ceiling `Check` before PRG (fast budget-exceeded denial), but defer `Consume` until after PRG passes. `Consume` now fails closed — a consume error denies with `BUDGET_ERROR`. This also narrows the check-to-consume window since `Consume` re-validates the ceiling.

Regression tests for no-consume-on-deny and fail-closed; updates the coverage test that asserted swallow-and-allow.

Note: opt-in `budget_id` resolution and fully atomic reserve semantics require a `BudgetGate` interface change and remain tracked in HELM-50.

Fixes HELM-50 (ordering + fail-closed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BggKqjoW6WCM93xa9VjdCx

PR #503, opened 2026-07-03
ci: fix TLA Specs — update rolling tla2tools.jar digest, fail download loudly

Hirama:
## Problem
Every TLA Specs run after 2026-07-03 18:24 UTC fails. Two causes:

1. **Stale digest**: tlaplus republished the v1.8.0 `tla2tools.jar` asset today (`updated_at: 2026-07-03T18:24:26Z`). v1.8.0 is a **rolling pre-release** — assets are rebuilt under the same tag — so the pinned SHA-256 (`237332bd…`) no longer matches the asset (`d144885f…`).
2. **Swallowed failure**: the Download TLC step runs the script inside `echo "jar=$(...)"`, which masks its exit code. On digest mismatch the job continued and later steps ran `java -cp "" tlc2.TLC` → misleading `ClassNotFoundException`.

## Fix
- Pin updated to `d144885f107ead1d1e56bcfc946df4b27b61588014b8d6525471c744aa743d18` in both the workflow and `scripts/tla/download-tools.sh`. Provenance: direct download from the official tlaplus v1.8.0 release; GitHub asset metadata confirms the republish by the tlaplus org.
- Download step now captures the script output separately and fails when the script fails or returns an empty path.
- Rolling-release caveat documented at both pin sites.

## Follow-up (not in this PR)
Vendor or mirror the jar to an immutable location — hash-pinning a rolling pre-release will break again on the next upstream rebuild.

Unblocks the `model-check` failures on PRs #496–#502.

### DIFF
--- a/.github/workflows/tla.yml
+++ b/.github/workflows/tla.yml
@@ -22,9 +22,20 @@ jobs:
       - name: Download TLC
         id: tlc
         env:
+          # NOTE: tlaplus v1.8.0 is a rolling pre-release; its tla2tools.jar is
+          # periodically rebuilt upstream, which changes this digest. When the
+          # job fails with a SHA-256 mismatch, re-verify the asset against the
+          # official tlaplus release and update the digest here and in
+          # scripts/tla/download-tools.sh together.
           TLA_TOOLS_VERSION: v1.8.0
-          TLA_TOOLS_SHA256: 237332bdcc79a35c7d26efa7b82c77c85c2744591c5598673a8a45085ff2a4fb
-        run: echo "jar=$(bash scripts/tla/download-tools.sh)" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


Hirama:
model-check is green on this branch — confirms the digest fix. @SergeyAP this unblocks the seven kernel fix PRs (#496–#502): they run the TLA workflow from their own branches, so once this merges I'll merge main back into them. Follow-up worth a ticket: vendor/mirror tla2tools.jar — v1.8.0 is a rolling pre-release, the pin will go stale again on the next upstream rebuild.

peycheff-com (approved):
Reviewed for the current quantum-docs unblock. Scope is limited to the TLA workflow/downloader digest and fail-loud classpath handling; all checks, including model-check, are green.

PR #508, opened 2026-07-04
cleanup(kernel): HELM-56 partial — IDs, PRG operator parity, clock warn

Hirama:
Addresses the safe, self-contained items from the HELM-56 cleanup batch. Riskier items deferred (see below).

## Changes
1. **Effect/lease IDs → crypto/rand** (`guardian`). Were minted from `clock.Now().UnixNano()`, which collides under same-nanosecond / high-throughput generation and leaks timing. Extracted the existing decision-ID pattern into a shared `randomID(prefix)` helper. `EffectID` is not part of the effect digest, so decision determinism is unaffected. Audit-log entry IDs left clock-derived on purpose (they are hashed into the deterministic audit chain — documented).
2. **PRG legacy `check()` operator parity** (`prg`). The artifact-only validator behind `Graph.Validate` handled only AND/OR; empty logic (`""`) and `NOT` fell through to a fail-safe deny, diverging from the canonical CEL path (`EvaluateRequirementSet`). Now handles `""` as AND and implements `NOT`, matching `combineResults`. Documented that `check()` stays artifact-only (no CEL).
3. **Warn on wall-clock fallback** (`guardian`). `NewGuardian` silently substituted `wallClock` when no `WithClock` was given, hiding a KERNEL_TCB §3 invariant gap. Now emits `slog.Warn`; behavior otherwise unchanged.

Each change has tests (randomID tight-loop collision test; Validate empty-logic + NOT tests). Full `guardian`+`prg` suites green under `-race`.

## Deferred from HELM-56 (with rationale)
- **`toMap` JSON round-trip / interceptor closures** — profiled the decision path: it is CPU-dominated by Ed25519 signing (~55% CPU). `toMap` is ~15% of *allocations* but <2% of wall time; closures don't register. Hand-building the CEL input map risks shape drift for negligible gain. Not worth the risk.
- **Remove ~20 deprecated `SetX` setters** — broad churn that breaks in-tree test callers for no behavior change. Needs a coordinated migration.
- **Hard-fail on missing authority clock** — breaks every caller relying on the default; warning added instead, hard-fail left for a coordinated migration.

Remaining HELM-56 scope stays open for the deferred items.


peycheff-com (approved):
Reviewed the HELM-56 partial cleanup. Focused on effect/lease ID determinism boundaries and PRG operator parity. Local validation passed in helm-ai-kernel-wt-pr508-review: go test ./core/pkg/guardian ./core/pkg/prg; make verify-boundary. GitHub required checks are green.

PR #518, opened 2026-07-06
test: cover npm dist-tag validation

mindburnlabs:
Follow-up to the npm dist-tag workflow repair that landed on main as 2256bd40.\n\nThis adds the missing unit coverage for the dist_tag validator:\n- default empty dist_tag resolves to latest\n- non-semver channel tags such as v0_7_backfill are allowed\n- SemVer and v-prefixed release tags are rejected\n- whitespace/shell-like tags are rejected\n\nLocal validation run:\n- python3 -m unittest validate_workflow_dispatch_inputs_test.py\n- validator CLI positive/negative publish dist_tag checks\n- git diff --check


peycheff-com (approved):
Code-owner approval for test-only follow-up covering npm dist-tag validator behavior.

PR #540, opened 2026-07-11
feat: add scoped emergency-stop fence

mindburnlabs:
## Summary

- adds an opt-in, tenant/workspace-scoped Kernel FENCE command and durable active-state barrier for new governed dispatches
- binds the FENCE_ACK signer identity, profile, and key material into the signed receipt hash; identity-changing replays fail closed
- prevents scope selection from untrusted JSON: evaluate and ext-authz bind configured scope, while the unauthenticated OpenAI proxy is unavailable when the fence is active
- pins and executes the merged contracts-catalog FENCE reference vectors (including Unicode, nanoseconds, six negative cases, and hybrid no-downgrade proof)

## Contract dependency

- contracts-catalog #25 is merged at a140cf32c9f245dc6ff2f1f97bbf64d1d2d82af0; the reference-pack manifest pins that exact source revision and content hash.

## Validation

- make lint
- make test
- make test-cli && make verify-fixtures
- make helm-chart-smoke
- go test -race -count=1 ./pkg/kernel ./pkg/guardian ./cmd/helm-ai-kernel
- strict Kernel/contract conformance review against the pinned catalog vector

## Explicit boundary

This is disabled by default and is only a Kernel-side fence for new governed dispatches. It is not a cross-plane operator Emergency Stop: Control Plane command ledger/outbox, ACK reconciliation, permit revocation, in-flight cancellation, adapter coverage, production configuration, and live evidence remain follow-on work.


peycheff-com (approved):
Approval for scoped emergency-stop fence: CI green, source-truth doc included, and public wording keeps it internal/opt-in rather than production Emergency Stop.

PR #561, opened 2026-07-14
feat(setup): harden project lifecycle and review evidence

mindburnlabs:
## Summary
- persist native lifecycle and recovery state inside a project-scoped namespace with secure locking and receipt handling
- harden migration, rollback, profile-drift, mode/signing, runtime-admission, and storage behavior
- add a public native-client boundary and maintainer lifecycle protocol; public claims now distinguish local setup, sterile client observation, and release/deployment evidence

## Validation
- go test -p 1 ./... -count=1
- go test -race ./pkg/store -count=1
- go vet ./...
- scripts/check_documentation_coverage.py
- scripts/check_documentation_truth.py

## Boundary
This change deliberately does not claim a real authenticated Codex or Claude Code session was observed. Local lifecycle evidence remains distinct from sterile-client observation and release authority.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ---
 title: HELM AI Kernel Changelog
-last_reviewed: 2026-07-10
+last_reviewed: 2026-07-14
 ---
 
 # Changelog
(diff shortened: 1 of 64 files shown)


mindburnlabs:
Superseded by reviewable stacked slices #562 (foundation) and #563 (recovery/evidence). The original exceeded the deterministic review-patch cap; these slices preserve the cap instead of bypassing it.

PR #569, opened 2026-07-15
feat(approval): bind source-owned grants

mindburnlabs:
## Summary

- define the first source-owned `ApprovalGrant` binding for HELM-142
- bind tenant, workspace, audience, pack/version/manifest, lifecycle action, canonical intent/effect/plan hashes, decision, policy version/epoch/hash, approval ceremony, signer set, server identity/trust key, issue/expiry window, and nonce
- add deterministic JCS sealing plus integrity and active-window validation
- cover field substitution, malformed hashes/nonces, unsupported decisions/actions, expiry boundaries, and deterministic resealing

## Security boundary

This PR defines a deterministic contract only. It does **not** authorize a mutation, verify a server or approver signature, consume replay state, persist a ceremony, activate a legacy approval route, or gate pack mutations. `ValidateAt` explicitly remains non-authorizing.

PR #549 remains the separate fail-closed containment record. This branch is based directly on `main` so the source-owned contract can be reviewed below the immutable reviewer patch cap.

HELM-142 remains Active after this PR. Follow-up slices must add trusted challenge/assertion verification, durable signer/quorum state, server-signed grant issuance carried into the existing ext-authz permit path, atomic single-use consumption, pack lifecycle enforcement, outcome/reconciliation, EvidencePack closure, and joined negative E2E evidence.

## Validation

- `GOWORK=off go test ./pkg/contracts -run 'TestApprovalGrant' -count=1`
- `GOWORK=off go test -race ./pkg/contracts -run 'TestApprovalGrant' -count=1`
- `GOWORK=off go test ./pkg/contracts -count=1`
- `GOWORK=off go vet ./pkg/contracts`
- `make verify-fixtures`
- `make verify-boundary`
- `make quality-pr`
- `git diff --check`

Linear: HELM-142

### DIFF
--- a/core/pkg/contracts/approval_grant.go
+++ b/core/pkg/contracts/approval_grant.go
@@ -1,0 +1,221 @@
+package contracts
+
+import (
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+	"unicode"
+)
+
+const (
+	ApprovalGrantSchemaV1   = "approval-grant.v1"
+	ApprovalGrantContractV1 = "2026-07-15"
+
+	ApprovalGrantDecisionAllow = "ALLOW"
+
+	ApprovalGrantActionInstall   = "install"
+	ApprovalGrantActionUpgrade   = "upgrade"
+	ApprovalGrantActionUninstall = "uninstall"
+	ApprovalGrantActionRollback  = "rollback"
+)
+
+var (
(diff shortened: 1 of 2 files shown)


mindburnlabs:
Superseded by #570 with an identical source tree and a DCO-compliant commit. The replacement preserves the published branch history instead of force-pushing it. No review or authorization from this draft should be carried forward implicitly; #570 must pass its own exact-head gates.

PR #581, opened 2026-07-16
fix(ci): fail closed contract-breaking gate (HELM-214)

mindburnlabs:
## Summary
- make contract-breaking comparisons fail closed when the exact remote base is unavailable
- remove the mutable PR-label override and verify the pinned oasdiff archive before extraction
- classify OpenAPI and Buf compatibility findings separately from tool failures, with hermetic regression coverage
- document the actual gate boundary and add the self-test to the applicable quality profiles

## Validation
- make test-contract-breaking
- GITHUB_BASE_REF=main make proto-breaking
- real OpenAPI gate in an ephemeral linux/amd64 container using the checksum-verified CI artifact
- make verify-boundary
- make docs-coverage docs-truth quality-self-test
- shellcheck and actionlint

## Boundary
This repairs the permit gate only. It does not approve a release, deployment, or the retained source-controlled major-version policy.


mindburnlabs:
## Validation context
- make test-contract-breaking
- GITHUB_BASE_REF=main make proto-breaking
- real OpenAPI gate in an ephemeral linux/amd64 container using the checksum-verified CI artifact
- make verify-boundary
- make docs-coverage docs-truth quality-self-test
- shellcheck and actionlint

Boundary: this repairs the permit gate only. It does not approve a release, deployment, or the retained source-controlled major-version policy.

mindburnlabs:
## Rebased onto `main` (v0.7.4) and pushed

`d7d53792` → **`82b1fa96`**, now a single commit on top of `a079cdfc` (`chore(release): v0.7.4 (#634)`). Force-with-lease, no bare force.

### What conflicted

Exactly one hunk: the single-line `.PHONY:` declaration at the top of `Makefile`. Both sides appended to it while `main` took 49 commits in 14 days.

Resolved as a **union**, not a preference: every token `main` added was kept, and this PR's one new token `test-contract-breaking` was inserted after `openapi-breaking`. Verified with a `comm` token-set diff — 0 `main`-only tokens lost, exactly 1 PR-only token added.

### What was dropped as already-in-main

**Nothing.** `main` has not implemented any part of this change — the approval (#588), connector-effect (#598), Codex sidecar (#556), launch-contract, release-permit and sandbox-sealing work all landed in unrelated paths. The PR is *not* redundant on inspection.

This was checked rather than assumed. The PR's own commit was isolated pre- and post-rebase (`d7d53792` vs `392eec71`, and `82b1fa96` vs `a079cdfc`) and the two patches diffed against each other. Raw diff-of-diffs: 3 lines, all expected (blob `index` lines, one hunk-header offset, the `.PHONY` line). Normalized diff-of-diffs is **byte-identical** — SHA-256 `8848da15dbe130943a28c8c4bda138e61247cc7aaa216ca46da8c9d9281deb9f` before and after.

### Behaviour change forced by the rebase

**None.** The rebase changed nothing beyond the lossless `.PHONY` union. No hunks collapsed into `main`, none reverted, no conflict markers anywhere in the tree, and all 9 changed files kept their original git file modes.

### What the PR itself changes (unchanged by the rebase, restated for review)

The headline is closing a **fail-open** path, not a style fix:

- **Unresolvable base silently passed the gate.** `git fetch … || true` plus a fallback to a local `$base_ref` meant that if `origin/main` could not be resolved, every spec hit `new on this branch (no base to diff) — skip` and the gate passed having compared nothing. Now an unavailable or malformed base is `exit 2`.
- **Mutable PR label could downgrade a break to a warning.** `CONTRACT_BREAKING_APPROVED`, fed from the `contract:breaking-approved` label, is removed. A source-controlled major-version bump is now the only compatibility escape. The label plumbing is gone from `ci.yml`; the only surviving references are in the new self-test, which sets the variable specifically to assert it *no longer* overrides.
- **Differ crashes were reported as contract breaks.** `$?` was never captured, so any `oasdiff`/`buf` failure (missing file, bad YAML, OOM) surfaced as "backward-incompatible contract change". Now `oasdiff` exit 1 and `buf` exit 100 are real findings (`exit 1`); every other non-zero exit is a tool failure (`exit 2`).
- **Missing `VERSION` in the base** defaulted `base_major` to `cur_major`, quietly making the major-bump escape unreachable. Now `exit 2`.
- The `broke` sentinel inversion in the openapi path is cosmetic — net semantics are unchanged apart from the deliberate removal of the override branch.

### Gates run and passed

| Gate | Result |
|---|---|
| `make test-contract-breaking` (this PR's own hermetic self-test) | pass |
| `make quality-self-test` | pass |
| `make openapi-breaking` | pass |
| `GITHUB_BASE_REF=main make proto-breaking` | pass |
| `shellcheck scripts/ci/{contract_breaking,test_contract_breaking}.sh` | pass |
| `actionlint .github/workflows/ci.yml` | pass |
| `make docs-coverage` | pass |
| `make docs-truth` | pass |
| `make verify-boundary` | pass |
| `make crucible` (L1 + L2) | pass — 12/12 use cases |
| `make verify-fixtures` | pass — all golden roots + all reference-pack vector suites, exact Go/Python parity |
| `make quantum-crypto-inventory` | pass |

`quantum-crypto-inventory` was not on the original gate list but its path globs (`.github/**`, `Makefile`, `scripts/**`) all cover this diff, so it was run additionally. L3 / A2A / OTel are **N/A** — zero files in those scopes are touched, so `crucible-full` was correctly not run. Reference-pack coverage is N/A; this is CI tooling only.

### Flagged for a human — please sanity-check these three

1. **`ci.yml` now hard-pins the `oasdiff` v1.23.0 tarball to a SHA-256** (`972b1053…`) and `sha256sum --check --strict`s it before extracting. Good supply-chain hygiene and arguably required — a gate you can't trust the differ of isn't a gate — but it is scope beyond "fail closed", and if upstream ever re-rolls that release asset CI breaks until the hash is bumped. Confirm you want the hard pin.
2. **No `quantum_posture` annotation was added.** Repo convention asks for one on crypto-touching changes. My read is that a SHA-256 integrity check on a CI download is not kernel crypto and does not change the classical Ed25519/SHA-256 posture or add any post-quantum control — and `quantum-crypto-inventory` passes. Confirm that read.
3. **`docs-truth` and the release-permit pair are not among the 18 required checks**, so their red is expected here and was not chased.

Pushed, gates green, **awaiting the Copilot round** that this push will spawn asynchronously. Not merging — that's Ivan's call.

peycheff-com (dismissed):

peycheff-com:
Superseded by #756. Current main retains and strengthens the fail-closed gate: no mutable label/env bypass, explicit base/tool failure handling, and immutable prior-tag release comparison.

PR #595, opened 2026-07-17
feat(kernel): add fenced dispatch admission

mindburnlabs:
## Summary

- add the signed `approval-dispatch-admission.v1` contract and pinned Ed25519 verification
- persist immutable tenant/workspace-scoped dispatch admissions under forced PostgreSQL RLS
- serialize new admission with FENCE and preserve exact replay/recovery as evidence
- expose separate dispatch-scoped internal claim and recovery routes
- document the exact pre-production authority boundary and startup role requirements

## Validation

- `GOWORK=off go test ./... -count=1`
- `make lint`
- `make verify-approval-ceremony-vectors`
- `make docs-truth`
- `make docs-coverage`
- real PostgreSQL forced-RLS race proof, including concurrent FENCE/admission ordering, response-loss replay, conflicting attempts, post-lock expiry, tamper detection, and tenant isolation

## Authority status

This is an internal, pre-production Kernel authority slice stacked on #594. It is not production-ready connector enforcement.

Release blockers intentionally remain:

- `connector_id` is asserted by the dispatch-scoped workload; source-owned policy/certification binding to the approved effect is not wired yet
- the data plane does not yet atomically bind a live signed admission to `CONSUMED -> DISPATCHING`
- FENCE does not yet enumerate active admissions or produce close/disposition and connector-boundary ACK evidence
- the runtime still performs idempotent DDL at startup; migration and DML-only runtime credentials are not split

Do not merge or describe this PR as end-to-end emergency-stop or production connector authority until those gates are implemented and proven.


mindburnlabs:
Fixed in f943eea1: registered both /internal/v1/approval-grants/admit-dispatch{,-recover} in RuntimeRouteSpecs() with RouteAuthWorkload + RouteRateKernel + RouteContractInternal (mirroring consume/recover), so dispatch-admission traffic gets the kernel/workload rate class instead of falling through to public.

PR #597, opened 2026-07-17
feat(kernel): persist certified connector release authority

mindburnlabs:
## Primary tracker

- HELM-74: R4 - First certified connector and EvidencePack path
- Prerequisite only for HELM-158 emergency-stop ordering; this PR does not close HELM-158.

## Summary

- define and verify source-owned signed connector release authority for an exact connector version and scope
- persist append-only certified/revoked history in PostgreSQL with exact-head succession, terminal revocation, forced RLS, and least-privilege writer/runtime roles
- use database time for planning-only current-certified reads; no API in this PR authorizes connector start
- harden trigger namespace resolution and prove temporary-table shadowing cannot bypass chain validation
- align Go, JSON Schema, Python, PostgreSQL, and JCS-safe revision bounds with 13 negative vectors

## Validation

- `cd core && GOWORK=off go test ./... -count=1`
- `cd core && GOWORK=off go test -race ./pkg/contracts ./pkg/registry/connectors -count=1`
- `HELM_TEST_POSTGRES_URL=postgres://... make test-connector-release-authority-postgres` (race detector, count 10)
- `make verify-fixtures`
- `make verify-boundary` (1000 current, 0 stale)
- `make lint`
- migration coverage, workflow YAML parse, docs coverage/truth, and `git diff --check`
- kernel guardian, connector conformance, and docs-truth preflight: GO for internal draft

## Truth boundary

Internal/pre-production only. This is a stacked draft on `codex/helm-158-connector-authority`. It is not production connector-start authority and proves no live deployment or irreversible effect.

Remaining production gates include the source signer/import and rotation path, approval BindingProvider wiring, durable idempotent effect admission with explicit not-started/started/uncertain outcomes, Data Plane integration, active-work reconciliation/evidence, and controlled deployment proof.

The new connector release contract is a source-owned cutover. Legacy ID-keyed connector metadata is intentionally not migrated because it lacks signed provenance and revision history.


mindburnlabs:
Fixed in 62206010: added VerifyStoredEnvelope (drops only the key.Enabled requirement; signature + pinned NotBefore/NotAfter still enforced) and routed validateStoredReleaseAuthority through it, so a key disabled/rotated after signing the current head can still read and revoke that stored head. New imports still require an enabled key. Regression test covers stored-verify-with-disabled-key (accept), new-import-with-disabled-key (reject), and tampered-stored-signature (reject). The kind-smoke failure was an infra flake (curl connection reset) — re-triggered.

PR #598, opened 2026-07-18
feat(kernel): reserve connector effects before dispatch

mindburnlabs:
## Summary

- bind approval authority to the exact connector action and certified release snapshot
- transactionally reserve an effect after FENCE/current-release checks and before connector dispatch
- persist the forced-RLS append-only lifecycle `ADMITTED -> STARTED -> NOT_STARTED|UNCERTAIN`
- make `STARTED` a single-winner claim and preserve reconciliation references
- recheck current FENCE and the exact certified release under the source-owned locks immediately before `STARTED`
- resolve a typed start denial as `NOT_STARTED` without issuing connector HTTP
- resolve connector-owned permit effect before MCP firewall and approval classification
- drive the real GitHub connector only after durable `STARTED`, with strict inputs and no mutating retry/redirect
- add real PostgreSQL concurrency, RLS, append-only, recovery, and admission/start versus FENCE/revocation ordering proof

## Stack

Depends on #597 (`codex/helm-158-connector-registry`). This PR contains four effect-admission commits only.

## Verification

- `HELM_TEST_POSTGRES_URL=... make test-effect-reservation-postgres` (`-race -count=10`)
- `GOWORK=off go test -race ./pkg/connectors/github ./pkg/runtimeadapters/mcp ./pkg/boundary/approvalceremony -count=1`
- `GOWORK=off go test ./... -count=1`
- `make docs-coverage docs-truth verify-boundary` (1000/1000 boundary entries)
- `make verify-approval-ceremony-vectors verify-connector-release-authority-vectors`
- `git diff --check`
- GitHub checks pending for commit `11b3eab3`

## Truth boundary

This is source-owned internal pre-production code and real-database proof. The path is opt-in and is not wired into a deployed HELM Cloud Data Plane. It does not establish live Emergency Stop, GA readiness, source-system acknowledgement, signed completion receipts, metering, or production operations.

## Remaining production gates

- deploy the approval binding and signed connector-release import services
- wire admission/recovery into the deployed Kernel and Data Plane
- add source acknowledgement plus signed outcome/receipt closure
- add FENCE/revocation active-work disposition and cross-plane reconciliation
- complete crash/restart, failover, load, abuse, SLO, and controlled live-effect evidence

PR #599, opened 2026-07-18
feat(kernel): close connector effects with signed evidence

mindburnlabs:
## Stack

Depends on #598. Review this PR as the signed effect-close slice only.

## What this adds

- connector-signed connector-effect-acknowledgement.v1 with exact issuer key connector release execution idempotency response and APPLIED or NOT_APPLIED bindings
- Kernel-signed effect-close-receipt.v1 bound to the exact prior reservation head and a verified EvidencePack
- atomic forced-RLS append-only approval_effect_closures plus COMPLETED lifecycle event
- exact replay conflicting replay and signature-validating recovery
- close from STARTED or UNCERTAIN including reconciliation after later FENCE or release revocation without granting new execution authority
- Go and independent Python reference vectors JSON Schemas CI and source-truth operations docs

## Local proof

- GOWORK=off go test ./... -count=1
- go test -race ./pkg/contracts ./pkg/boundary/approvalceremony -count=1
- HELM_TEST_POSTGRES_URL=postgres://... make test-effect-reservation-postgres (race count 10 on the pinned CI PostgreSQL image)
- make verify-approval-ceremony-vectors verify-connector-release-authority-vectors verify-effect-close-vectors
- make verify-boundary
- workflow YAML and JSON schema parse checks
- git diff --check

## Truth boundary

This is internal pre-production Kernel contract persistence and test evidence. It does not deploy a Data Plane close endpoint make connectors publish acknowledgements or EvidencePacks implement active-work disposition or compensation or prove controlled-live HELM Cloud behavior. Keep this PR draft and stacked until its parent chain and later deployment gates are source-authorized.


mindburnlabs:
Fixed in b1004859: moved the strict enabled-key VerifyEnvelope into closeEffectReservation's new-close (Started/Uncertain) branch, and use VerifyStoredEnvelope for the service preflight + the store preflight. So an exact retry of an already-COMPLETED close now returns the stored receipt via the replay path even after the connector ack key was disabled, while a genuinely new close still requires an enabled key. Signature + pinned lifetime enforced throughout; effect-close vectors (2 sig, 7 negative mutations) still verify with exact Go/Python parity.

PR #600, opened 2026-07-18
feat(kernel): record signed active-work dispositions

mindburnlabs:
## Summary

- add Control Plane-signed active-work disposition commands bound to the exact tenant/workspace, current FENCE, reservation head, connector identity, and append-only predecessor
- persist forced-RLS append-only disposition history and return Kernel-signed receipts with `execution_authority: NONE`
- require active-FENCE close to bind the latest current-FENCE non-`HOLD` disposition; preserve immutable historical close recovery after later FENCE
- bind connector acknowledgement and terminal close receipt to the exact disposition receipt
- add source-owned JSON Schemas, deterministic Go fixtures, independent Python verification, operator docs, and CI gates

## Security and ordering

- `HOLD` blocks terminal close
- no disposition under active FENCE fails closed
- a pre-rotation disposition fails after a newer FENCE epoch
- forged disposition rows are rejected during successor append and close by full signature and authority verification
- wrong-scope direct insert plus update/delete mutation attempts are rejected
- cancellation and compensation requests remain intent only and grant no external execution authority

## Verification

- `GOWORK=off go test ./... -count=1`
- `GOWORK=off go vet ./...`
- `GOWORK=off go test -race ./pkg/contracts ./pkg/boundary/approvalceremony -count=1`
- pinned real PostgreSQL integration with `-race -count=10`
- `make verify-fixtures`
- `make docs-coverage`
- `make docs-truth`
- `make verify-boundary`
- YAML, JSON, canonical vector, and diff hygiene checks

## PR preflight

- Kernel guardian / high-risk loop: GO after resolving current-FENCE close bypass
- conformance engineer: GO after adding unit-gate and direct INSERT/DELETE proofs
- docs truth: GO; claims match code and remain explicitly pre-production

## Delivery boundary

This is internal pre-production Kernel contract and PostgreSQL evidence. It does not provide a deployed Control Plane command ledger/outbox, authenticated Data Plane disposition API, connector reconciliation/cancel/compensate adapter, KMS/HSM deployment, controlled-live source proof, Emergency Stop release authority, or GA authority.

Stacked on #599. Keep draft and do not merge until the base stack and source-owned release gates are satisfied.


mindburnlabs:
Fixed in a553e172: effectAcknowledgementMatchesDisposition now takes the close-time reservation and requires the disposition command's reservation_sequence, reservation_head_hash, and reservation_state to equal the reservation being closed (both the new-close path passing `current` and the recover/replay path passing `prior`). A disposition recorded while STARTED can no longer bind an old receipt and close a different active head after a same-fence STARTED→UNCERTAIN transition. Disposition vectors (2 sig, 7 negative mutations) still verify with exact Go/Python parity. Getting a kernel-guardian review of this auth-binding change in parallel with CI.

PR #603, opened 2026-07-18
ci(release): publish a signed image for an arbitrary commit sha (HELM-129)

SergeyAP:
## What

Adds a `workflow_dispatch` path to `release.yml` that publishes a signed image
for an **arbitrary commit sha** (`container-sha`), so `qa-deploy` can pull a
registry image for a PR head instead of side-loading one into the cluster.

Mirrors the `source_sha` dispatch already used by
`svc-helm-control-plane`'s `release-oci.yml`, so the two repos behave the same
way from `qa-deploy`'s perspective.

The `v*` tag pipeline is untouched. The two paths are mutually exclusive by
construction:

- `version-contract` gains `if: github.event_name == 'push'`. It is the root of
  the tag pipeline and every other release job `needs`-chains from it, so on a
  dispatch the whole `v*` pipeline skips.
- `container-sha` carries `if: github.event_name == 'workflow_dispatch'`, so it
  never runs on a tag push.

There are no `always()` / `!cancelled()` jobs in `release.yml`, so the skip
chain is clean — nothing downstream runs on a dispatch by accident.

## The security fix worth reviewing carefully

The job gates on "this commit has green CI" before it builds and signs. The
naive form of that gate is exploitable:

```
gh api "/repos/$REPO/actions/workflows/ci.yml/runs?head_sha=$SHA&status=success" | jq '.total_count'
```

`ci.yml` triggers on `pull_request`:

```
$ sed -n '1,8p' .github/workflows/ci.yml
name: CI
on:
  workflow_dispatch:
  pull_request:
    branches: [main]
  push:
    branches: [main]
```

so a **fork's** PR run also lands in this workflow's run list, carrying the
fork's `head_sha`. An unfiltered `total_count` would therefore accept a sha
whose only green CI run came from untrusted, never-merged fork code — and then
build it, push it to `ghcr.io/mindburn-labs/helm-ai-kernel`, and **cosign-sign
it under the release identity**. The cosign identity of a dispatch build is the
same `release.yml` as a real `v*` release, so a consumer verifying the
signature could not tell the difference.

Fixed by requiring the run to originate in this repo:

```
matching="$(gh api \
  "/repos/${GITHUB_REPOSITORY}/actions/workflows/ci.yml/runs?head_sha=${SOURCE_SHA}&status=success&per_page=100" \
  | jq --arg repo "${GITHUB_REPOSITORY}" \
      '[.workflow_runs[] | select(.head_repository.full_name == $repo)] | length')"
```

Fork PR runs have `head_repository.full_name` = the fork, so they are excluded.

## Other guards

- **Input validation.** `source_sha` must match `^[0-9a-f]{40}$` — a full
  lowercase sha, rejected before checkout.
- **Concurrency.** `group: container-sha-${{ inputs.source_sha }}` with
  `cancel-in-progress: false`. The tag is deterministic, so two concurrent
  dispatches for the same sha would otherwise race the same ghcr tag to two
  different digests.
- **Least privilege.** The job declares its own `permissions` block
  (`actions: read`, `contents: read`, `id-token: write`, `packages: write`)
  rather than widening the workflow default of `read-all`.

## Image identity — the build is labelled dev-grade

Because a dispatch build is signed by the same workflow as a release, the image
has to state what it is. Version is an explicit prerelease
(`0.0.0-sha.<short12>`), never a released semver, and the labels say so:

```
dev.mindburn.build-grade=dev
dev.mindburn.build-origin=release.yml/container-sha(workflow_dispatch)
org.opencontainers.image.revision=<source_sha>
```

## Tag format

The image is tagged `sha-<full 40-char sha>`, matching what `qa-deploy`
resolves and pulls — the skill takes `headRefOid` for a PR and
`git rev-parse` for a tag/commit, both full shas, then checks
`manifests/sha-$sha`. Verification therefore uses the **full** sha:

```
crane config ghcr.io/mindburn-labs/helm-ai-kernel:sha-<full-40-char-sha> | jq '.config.Labels'
```

(A short sha will 404 — worth stating, because an earlier draft of the internal
runbook said `sha-<short>` and that reads as a broken release.)

## First real run

Dispatch against a sha from **merged `main`**, not a PR head, so the first
exercise of the path is on a commit that already satisfies the fork filter the
normal way. Then confirm the labels above are present on the pushed image.

## Scope note

The same fork-filter weakness exists in the reference implementation this
mirrors (`svc-helm-control-plane`'s `release-oci.yml`). That backport is tracked
separately as HELM-255, and its first step is deliberately to check the
premise — whether that repo's `ci.yml` triggers on `pull_request` at all. It was
confirmed here for `helm-ai-kernel` (output above); it has **not** been checked
for the control plane. If that trigger is absent, the hole is unreachable there
today and should be recorded and downgraded rather than fixed blind.

Refs HELM-129. Related: HELM-255.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -317,6 +317,11 @@ jobs:
     # gate) and this job skips on tag pushes.
     runs-on: ubuntu-latest
     if: github.event_name == 'workflow_dispatch'
+    # One publish in flight per sha: the tag is deterministic, so two
+    # concurrent dispatches would race the same ghcr tag to different digests.
+    concurrency:
+      group: container-sha-${{ inputs.source_sha }}
+      cancel-in-progress: false
     permissions:
       actions: read
       contents: read
@@ -337,11 +342,19 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           set -euo pipefail
-          total="$(gh api "/repos/${GITHUB_REPOSITORY}/actions/workflows/ci.yml/runs?head_sha=${SOURCE_SHA}&status=success" --jq '.total_count')"
-          if [[ "${total}" -lt 1 ]]; then
-            echo "::error::No successful CI (ci.yml) run found for ${SOURCE_SHA}; refusing to publish an unverified commit"
+          # Only count runs whose head repository is THIS repo. A pull_request
+          # run from a fork also lands in this workflow's run list with the
+          # fork's head_sha, so an unfiltered total_count would let a dispatch
+          # publish a signed image built from untrusted, never-merged fork code.
+          matching="$(gh api \
+            "/repos/${GITHUB_REPOSITORY}/actions/workflows/ci.yml/runs?head_sha=${SOURCE_SHA}&status=success&per_page=100" \
+            | jq --arg repo "${GITHUB_REPOSITORY}" \
+                '[.workflow_runs[] | select(.head_repository.full_name == $repo)] | length')"
+          if [[ "${matching}" -lt 1 ]]; then
+            echo "::error::No successful CI (ci.yml) run for ${SOURCE_SHA} originating in ${GITHUB_REPOSITORY}; refusing to publish an unverified commit"
             exit 1
           fi
+          echo "Found ${matching} successful in-repo CI run(s) for ${SOURCE_SHA}"
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
         with:
           ref: ${{ inputs.source_sha }}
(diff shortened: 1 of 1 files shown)


SergeyAP:
Superseded by #604, which is merged (`82d384c1`).

This branch carried identical content but could not pass `Signed-off-by check`:
all three of its commits, including the `main` merge commit, lacked DCO trailers.

```
##[error]Commit fff070d0951239b6500626f1edcc6b10a728e7d0 is missing a Signed-off-by trailer
##[error]Commit c77dea08a9aa6a9177ac4bec7cefa0502bdbdf6d is missing a Signed-off-by trailer
##[error]Commit 563cc03e492527a05e2e03a724a4951286267e09 is missing a Signed-off-by trailer
```

The usual remedy — `git commit --amend -s` plus a force-push — rewrites history,
so instead #604 was cut fresh from `main` with the two commits cherry-picked
using `-s`. That also removed the merge commit entirely, leaving nothing else to
sign. `release.yml` was byte-identical between the two branches.

Closing in favour of #604. No content was lost.

PR #574, opened 2026-07-16
ci: contract breaking-change gate on PRs (HELM-151 GATE 1)

SergeyAP:
## What & why

GATE 1 of the HELM-149 two-gate compatibility model, for the kernel (the richest contract surface). A PR that **backward-breaks a contract** should fail CI, so a break can't ship silently under a "minor" version — that's what makes the semver work of HELM-146/147 mean something.

**Gap this closes:** today the kernel checks breaking changes **only at release** (`proto-breaking` lives in the `release`/`nightly` profiles; the `merge` profile that also has it is not run by any workflow). On a **PR** there is no breaking check at all, and **OpenAPI has no breaking gate anywhere** (only `sdk-openapi-check`, which is SDK↔spec drift, not a backward-compat diff).

## Change

- **`scripts/ci/contract_breaking.sh`** — diffs a surface against the PR's base branch:
  - OpenAPI (`api/openapi/helm.openapi.yaml`, `protocols/specs/effects/openapi.yaml`) via `oasdiff breaking --fail-on ERR` (oasdiff prints but exits 0 by default — `--fail-on ERR` makes it fail).
  - proto (`protocols/policy-schema`) via `buf breaking --against <base>`.
  - Two **explicit, visible** escapes: a **major-version bump** (the break is intended and signalled by the version), and the **`contract:breaking-approved` PR label** (`CONTRACT_BREAKING_APPROVED=1` → break downgraded to a `::warning::` for deliberate pre-1.0 churn).
  - Portable: works on macOS bash 3.2 and needs no PyYAML (version parsed with awk); `shellcheck`-clean.
- **`Makefile`** — `openapi-breaking` (new) + `proto-breaking` routed through the wrapper (uniform escape/override).
- **`scripts/ci/quality-gates.json`** — new `openapi-breaking` gate; both breaking gates added to the **`pr` profile** (previously absent). `make quality-self-test` passes (45 gates, 11 profiles).
- **`.github/workflows/ci.yml`** — the `quality-pr` job now sets up `buf`, installs `oasdiff` (`go install …@v1.23.0`), and passes the label override env.

Because the gates hang off the existing `quality.py` profile runner (impact-filtered by `paths`), no parallel pipeline is introduced — the gate only runs when a contract file changes.

## Verified locally

| Case | Result |
| --- | --- |
| No contract change | pass (rc 0) |
| Removed endpoint (`/api/health`), no escape | **fail (rc 1)** + `::error::` |
[...]

PR #560, opened 2026-07-14
docs: separate Kernel from HELM product motions

mindburnlabs:
## Summary
- Makes explicit that Company Builder, Company OS, Network, external builders, social surfaces, and financial rails are outside the OSS Kernel feature boundary.
- companion to Mindburn-Labs/docs#74

## Validation
- make docs-coverage docs-truth: PASS
- diff check: PASS

## Truth boundary
Target product motions remain distinct from current implementation, release, revenue, and public-claim truth.

PR #559, opened 2026-07-14
docs(changelog): complete v0.7.2 notes for shipped enforcement + deps

SergeyAP:
## Why this PR is needed

The `v0.7.2` tag captures **13 commits since `v0.7.1`**, but the current 0.7.2
changelog block only documents two of them (the Go 1.25.12 rebuild and the
`BoundaryStatus` OpenAPI alignment). Several **developer-visible** changes ship
in the same tag with no changelog line. Since the tag is the public release
record, this PR completes the 0.7.2 notes before it is cut.

This is documentation only — no code, surface, or version change.

## What is added to the 0.7.2 block

| Ref | Change | Why it belongs in the notes |
| --- | --- | --- |
| #534 | Governed MCP execution bridge — real `ALLOW`/`DENY`/`ESCALATE` verdicts, replacing the deny-only adapter | New developer-visible enforcement path; fail-closed by default |
| #540 | Scoped emergency-stop fence (with recorded cryptography posture; JCS Unicode separators preserved) | New containment feature |
| #533 | Replay-determinism verifier gate now fails closed | Verifier behavior change |
| #532 | Sandbox fails closed on ambiguous network posture | Sandbox behavior change |
| #522, #539, #538 | `golang.org/x/crypto`, Jackson Databind, and Python `cryptography` build dependency updates | Security-relevant dependency hardening |

The `quantum_posture` annotation is widened to match the broader scope while
keeping its assertion true — **none** of these add a post-quantum cryptographic
control.

## Wording discipline

Descriptions are source-backed against the referenced commits and add **no new
capability, GA, or CVE-clearance claim** — matching the house style applied to
this block on merge of #550 (CVE clearance stays evidence-gated post-release).

## Validation

```
$ make docs-truth
Documentation coverage check passed: 651 rows cover 383 active surfaces.
Documentation truth check passed: 651 coverage rows resolve to live sources and docs.
$ python3 scripts/release/check_version_drift.py local --tag v0.7.2
drift: PASS
```

## Sequencing

This must merge **before** the `v0.7.2` tag is pushed, so the tagged release
carries the complete notes. No version surfaces or VEX change here — those are
already on `main` (#536, #550).

Refs **HELM-133**.

PR #550, opened 2026-07-12
release: finalize v0.7.2 changelog and OpenVEX

SergeyAP:
## Purpose

This PR adds the two source-controlled **pre-tag assets** required for the
`v0.7.2` release train:

1. A dated `0.7.2` block in `CHANGELOG.md`, moving the already-staged
   `BoundaryStatus` notes out of `[Unreleased]`.
2. `release/vex/v0.7.2.openvex.json`, which `stage_release_assets.sh` requires
   when the future `v0.7.2` tag triggers `release.yml`.

It does **not** create a tag, publish images, package a chart, or publish a
GitHub release. It also does not authorize the release to be cut; the broader
release train still needs its independent security, review, and release gates.

## Security-claim boundary

`main` already pins Go 1.25.12. The upstream Go release history records
cumulative standard-library security fixes in `crypto/x509`, `mime`,
`net/textproto`, `crypto/tls`, and `os`. Once an approved `v0.7.2` tag is
created, the release workflow will rebuild the two container images using that
toolchain.

This PR makes **no claim that a CVE is already cleared**: no `v0.7.2` image has
been built or scanned yet. Post-release Trivy, Artifact Hub, and published
artifact evidence are required before asserting any CVE clearance or a changed
Artifact Hub grade. `GO-2026-5932` (`x/crypto/openpgp`) remains a separate
tracking concern.

## Scope

- No version-surface changes; `main` is already at `0.7.2`.
- No toolchain changes; `main` already pins Go 1.25.12.
- No runtime, product, or cryptographic behavior change.

## Validation

```text
/usr/bin/python3 scripts/release/check_version_drift.py local
Documentation truth check passed: 651 coverage rows resolve to live sources and docs.
git diff --check
```

## Release sequence after approval and merge

1. Complete the remaining independent security and release approvals.
2. Merge this metadata PR to `main`.
3. Run the approved release process from final `main`, including tag creation.
4. Collect release receipts and image-scanning evidence before making public
   CVE-clearance or Artifact-Hub-grade claims.

PR #604, opened 2026-07-18
ci(release): publish a signed image for an arbitrary commit sha (HELM-129)

SergeyAP:
## What

Adds a `workflow_dispatch` path to `release.yml` that publishes a signed image
for an **arbitrary commit sha** (`container-sha`), so `qa-deploy` can pull a
registry image for a PR head instead of side-loading one into the cluster.

Mirrors the `source_sha` dispatch already used by
`svc-helm-control-plane`'s `release-oci.yml`, so the two repos behave the same
way from `qa-deploy`'s perspective.

The `v*` tag pipeline is untouched. The two paths are mutually exclusive by
construction:

- `version-contract` gains `if: github.event_name == 'push'`. It is the root of
  the tag pipeline and every other release job `needs`-chains from it, so on a
  dispatch the whole `v*` pipeline skips.
- `container-sha` carries `if: github.event_name == 'workflow_dispatch'`, so it
  never runs on a tag push.

There are no `always()` / `!cancelled()` jobs in `release.yml`, so the skip
chain is clean — nothing downstream runs on a dispatch by accident.

## The security fix worth reviewing carefully

The job gates on "this commit has green CI" before it builds and signs. The
naive form of that gate is exploitable:

```
gh api "/repos/$REPO/actions/workflows/ci.yml/runs?head_sha=$SHA&status=success" | jq '.total_count'
```

`ci.yml` triggers on `pull_request`:

```
$ sed -n '1,8p' .github/workflows/ci.yml
name: CI
on:
  workflow_dispatch:
  pull_request:
    branches: [main]
  push:
    branches: [main]
```

so a **fork's** PR run also lands in this workflow's run list, carrying the
fork's `head_sha`. An unfiltered `total_count` would therefore accept a sha
whose only green CI run came from untrusted, never-merged fork code — and then
build it, push it to `ghcr.io/mindburn-labs/helm-ai-kernel`, and **cosign-sign
it under the release identity**. The cosign identity of a dispatch build is the
same `release.yml` as a real `v*` release, so a consumer verifying the
signature could not tell the difference.

Fixed by requiring the run to originate in this repo:

```
matching="$(gh api \
  "/repos/${GITHUB_REPOSITORY}/actions/workflows/ci.yml/runs?head_sha=${SOURCE_SHA}&status=success&per_page=100" \
  | jq --arg repo "${GITHUB_REPOSITORY}" \
      '[.workflow_runs[] | select(.head_repository.full_name == $repo)] | length')"
```

Fork PR runs have `head_repository.full_name` = the fork, so they are excluded.

## Other guards

- **Input validation.** `source_sha` must match `^[0-9a-f]{40}$` — a full
  lowercase sha, rejected before checkout.
- **Concurrency.** `group: container-sha-${{ inputs.source_sha }}` with
  `cancel-in-progress: false`. The tag is deterministic, so two concurrent
  dispatches for the same sha would otherwise race the same ghcr tag to two
  different digests.
- **Least privilege.** The job declares its own `permissions` block
  (`actions: read`, `contents: read`, `id-token: write`, `packages: write`)
  rather than widening the workflow default of `read-all`.

## Image identity — the build is labelled dev-grade

Because a dispatch build is signed by the same workflow as a release, the image
has to state what it is. Version is an explicit prerelease
(`0.0.0-sha.<short12>`), never a released semver, and the labels say so:

```
dev.mindburn.build-grade=dev
dev.mindburn.build-origin=release.yml/container-sha(workflow_dispatch)
org.opencontainers.image.revision=<source_sha>
```

## Tag format

The image is tagged `sha-<full 40-char sha>`, matching what `qa-deploy`
resolves and pulls — the skill takes `headRefOid` for a PR and
`git rev-parse` for a tag/commit, both full shas, then checks
`manifests/sha-$sha`. Verification therefore uses the **full** sha:

```
crane config ghcr.io/mindburn-labs/helm-ai-kernel:sha-<full-40-char-sha> | jq '.config.Labels'
```

(A short sha will 404 — worth stating, because an earlier draft of the internal
runbook said `sha-<short>` and that reads as a broken release.)

## First real run

Dispatch against a sha from **merged `main`**, not a PR head, so the first
exercise of the path is on a commit that already satisfies the fork filter the
normal way. Then confirm the labels above are present on the pushed image.

## Scope note

The same fork-filter weakness exists in the reference implementation this
mirrors (`svc-helm-control-plane`'s `release-oci.yml`). That backport is tracked
separately as HELM-255, and its first step is deliberately to check the
premise — whether that repo's `ci.yml` triggers on `pull_request` at all. It was
confirmed here for `helm-ai-kernel` (output above); it has **not** been checked
for the control plane. If that trigger is absent, the hole is unreachable there
today and should be recorded and downgraded rather than fixed blind.

## Supersedes #603

#603 carried identical content but failed the `Signed-off-by check` — all three
of its commits (including the `main` merge commit) lacked DCO trailers. Rather
than rewrite that branch, this one is cut fresh from current `main` and the two
commits are cherry-picked with `-s`, so there is no merge commit to sign either.

`release.yml` is byte-identical between the two branches:

```
$ diff <(git show HEAD:.github/workflows/release.yml) \
       <(git -C ../helm-ai-kernel-helm-129 show fff070d0:.github/workflows/release.yml)
# no output — identical
```

DCO now passes under the same regex `dco.yml` uses:

```
$ for sha in $(git rev-list origin/main..HEAD); do
    git log -1 --format=%B "$sha" | grep -Eqi '^Signed-off-by: .+ <.+@.+>' && echo "$sha OK"; done
ad0177db1316e637e02097ee704c7ccefee2653b OK
1a88693857bb69daa402f97df21619a2309b66af OK
```

Refs HELM-129. Related: HELM-255.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -317,6 +317,11 @@ jobs:
     # gate) and this job skips on tag pushes.
     runs-on: ubuntu-latest
     if: github.event_name == 'workflow_dispatch'
+    # One publish in flight per sha: the tag is deterministic, so two
+    # concurrent dispatches would race the same ghcr tag to different digests.
+    concurrency:
+      group: container-sha-${{ inputs.source_sha }}
+      cancel-in-progress: false
     permissions:
       actions: read
       contents: read
@@ -337,11 +342,19 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           set -euo pipefail
-          total="$(gh api "/repos/${GITHUB_REPOSITORY}/actions/workflows/ci.yml/runs?head_sha=${SOURCE_SHA}&status=success" --jq '.total_count')"
-          if [[ "${total}" -lt 1 ]]; then
-            echo "::error::No successful CI (ci.yml) run found for ${SOURCE_SHA}; refusing to publish an unverified commit"
+          # Only count runs whose head repository is THIS repo. A pull_request
+          # run from a fork also lands in this workflow's run list with the
+          # fork's head_sha, so an unfiltered total_count would let a dispatch
+          # publish a signed image built from untrusted, never-merged fork code.
+          matching="$(gh api \
+            "/repos/${GITHUB_REPOSITORY}/actions/workflows/ci.yml/runs?head_sha=${SOURCE_SHA}&status=success&per_page=100" \
+            | jq --arg repo "${GITHUB_REPOSITORY}" \
+                '[.workflow_runs[] | select(.head_repository.full_name == $repo)] | length')"
+          if [[ "${matching}" -lt 1 ]]; then
+            echo "::error::No successful CI (ci.yml) run for ${SOURCE_SHA} originating in ${GITHUB_REPOSITORY}; refusing to publish an unverified commit"
             exit 1
           fi
+          echo "Found ${matching} successful in-repo CI run(s) for ${SOURCE_SHA}"
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
         with:
           ref: ${{ inputs.source_sha }}
@@ -378,6 +391,16 @@ jobs:
             BUILD_TIME=${{ steps.build-meta.outputs.build_time }}
           tags: |
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:sha-${{ inputs.source_sha }}
+          # The cosign identity of a dispatch build is the same release.yml as
+          # a v* release, so the image itself must state that it is dev-grade:
+          # never a released version, only a preview of one commit.
+          labels: |
+            org.opencontainers.image.source=https://github.com/${{ github.repository }}
+            org.opencontainers.image.revision=${{ inputs.source_sha }}
+            org.opencontainers.image.version=${{ steps.build-meta.outputs.version }}
+            org.opencontainers.image.created=${{ steps.build-meta.outputs.build_time }}
+            dev.mindburn.build-grade=dev
+            dev.mindburn.build-origin=release.yml/container-sha(workflow_dispatch)
       - uses: sigstore/cosign-installer@398d4b0eeef1380460a10c8013a76f728fb906ac # v3
       - name: Sign sha image keyless
         env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,15 @@ name: Release
 on:
   push:
     tags: ["v*"]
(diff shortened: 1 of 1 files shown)


Hirama (approved):

SergeyAP:
## CI status: 34 pass, 1 fail — and the failure predates this branch

The only red check is **HELM Autonomous Release Permit**, failing at the step
`Reduce distinct-provider reviews`. Evidence that it is not a regression from
this PR:

**1. It fails identically on merged PRs.** The four most recent merged PRs all
carry the same failure:

| PR | Autonomous Release Permit | merged |
| --- | --- | --- |
| #601 | fail | 2026-07-18 13:35 UTC |
| #574 | fail | merged |
| #560 | fail | merged |
| #559 | fail | merged |
| #550 and older | check not present | merged |

So the gate was introduced recently and has been red on every PR since. #601
touched the kernel principal↔tenant binding registry — nothing to do with
release workflows — and failed the same way.

**2. It is not a required check.** Branch protection is off on this repo:

```
$ gh api repos/Mindburn-Labs/helm-ai-kernel/branches/main/protection
{"message":"Branch protection has been disabled on this repository.", "status":"404"}
```

**3. The failing step is about reviews, not about this diff.** The job's earlier
steps all succeed — permit context downloaded, authority generation verified,
source-owned permit verifier built. It fails only when reducing *distinct-provider
reviews*, i.e. it wants independent reviews that a freshly opened PR does not yet
have. Nothing in it inspects the contents of `release.yml`.

**4. Everything that does inspect this diff is green** — `Deterministic
repository gates`, `contract-drift`, `kernel`, `kernel-coverage`, `kind-smoke`,
`deployment-smoke`, `release-smoke`, `local-validation`, CodeQL across all
languages, `Rust audit`, `Coverage and truth`, and `Signed-off-by check`.

### One thing I could not establish

I could not find this gate's definition in the repository. It reports its path
as `.github/workflows/ci.yml`, but `ci.yml` on this branch and on `origin/main`
contains no such job, and nothing under `.github/` matches `distinct-provider`
or `Autonomous Release Permit`. Its workflow id also 404s:

```
$ gh run view --job 88115816976 --log
failed to get run: HTTP 404 (…/actions/workflows/312927024)
```

So I can show that the check is red everywhere and blocks nothing, but I cannot
show what it would take to make it green, or confirm that going ahead is the
intended behaviour rather than a governance gate that is silently broken. That
call belongs to whoever owns the release-authority surface — flagging it rather
than assuming it is noise.

PR #637, opened 2026-07-21
fix(prg): refuse rules Graph.Validate cannot evaluate instead of passing them

mindburnlabs:
## Problem

`check` in `core/pkg/prg/graph.go` evaluates **artifact presence only**. A requirement carrying a CEL `Expression` but no `ArtifactType` fell into this branch:

```go
} else {
    // Expression or others ignored in this legacy check
    has = true
}
```

So `Graph.Validate` could return `true` **and `rule.Hash()`** for a policy whose expression was never evaluated — a satisfied verdict carrying evidence of an evaluation that did not happen.

That is a fail-open path inside a component whose stated posture is fail-closed.

## Fix

`Validate` now refuses rules it cannot faithfully evaluate, and names the evaluator that can — which is what the function's own doc comment already directed callers toward:

> The canonical evaluator that honours CEL is `PolicyEngine.EvaluateRequirementSet` — prefer it for policy decisions.

The malformed case (neither `ArtifactType` nor `Expression`) now fails closed instead of counting as satisfied.

**Why refuse rather than silently return `false`:** a caller reaching this path has picked the wrong evaluator. An error says so; a `false` would look like a policy denial and send someone debugging the wrong thing.

## Blast radius

Effectively zero today. `Graph.Validate` has **no production callers** — only `pkg/prg/engine_comprehensive_test.go` — and the guardian decision path uses `PolicyEngine.EvaluateRequirementSet` (`guardian.go:477`). This closes an exported foot-gun *before* something calls it.

## Tests

Three new cases, six existing ones unchanged:

```
--- PASS: TestGraph_Validate_RefusesCELExpressionRequirement
--- PASS: TestGraph_Validate_RefusesNestedCELExpressionRequirement   (child sets, not just top level)
--- PASS: TestGraph_Validate_MalformedRequirementFailsClosed
```

The first asserts the full fail-open signature: not satisfied, **and** no rule hash emitted.

## Boundary

`core/pkg/prg` has **0 entries** in `tools/boundary/protected.manifest` (guardian has 41), so no manifest regeneration and no `[KERNEL-SYNC]` follow-up is required. Confirmed rather than assumed:

```
make verify-boundary → Results: 1051 current, 0 stale, 0 errors → OSS boundary check passed. ✓
GOWORK=off go vet ./pkg/prg/  → clean
GOWORK=off go test ./pkg/prg/ → ok
```

## Provenance

Found by a full 318-issue audit of the HELM AI board against source-owned evidence. **This path is named in no existing Linear issue** — it surfaced while surveying the HELM-56 remainder. It needs an issue filed; Linear writes are currently held pending an MCP re-auth.

A sibling fail-open found in the same survey — `guardian.go:468` discarding the `toMap` error, so CEL evaluates against an empty `input.effect` — is **not** in this PR: it sits in the protected guardian package and is batched with the HELM-52/56 work.

### DIFF
--- a/core/pkg/prg/engine_comprehensive_test.go
+++ b/core/pkg/prg/engine_comprehensive_test.go
@@ -929,3 +929,75 @@ func TestCompiler_Compile_ContainsRule(t *testing.T) {
 		t.Fatal("compiled graph should contain rule with ID as key")
 	}
 }
+
+// Graph.Validate evaluates artifact presence only. Before this guard, a
+// requirement carrying a CEL Expression but no ArtifactType was assumed
+// satisfied, so Validate could return true *plus a rule hash* for a policy
+// whose expression was never evaluated — a fail-open verdict carrying evidence
+// of an evaluation that did not happen.
+func TestGraph_Validate_RefusesCELExpressionRequirement(t *testing.T) {
+	g := NewGraph()
+	_ = g.AddRule("act-cel", RequirementSet{
+		ID:    "rs-cel",
+		Logic: AND,
+		Requirements: []Requirement{
+			{ID: "r1", Expression: "intent.risk < 3"},
+		},
+	})
+
+	ok, hash, err := g.Validate("act-cel", nil)
+	if err == nil {
+		t.Fatal("expected Validate to refuse a CEL expression requirement")
+	}
+	if ok {
+		t.Fatal("fail-open: reported satisfied without evaluating the expression")
+	}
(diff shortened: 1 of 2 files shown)


mindburnlabs:
Superseded by #651 — same change with a DCO `Signed-off-by` trailer (the check rejected the unsigned commit here, and force-push is blocked in this workspace so it had to be a fresh branch).

PR #638, opened 2026-07-21
fix(guardian): enforce tainted-egress denial by default, reserve the override key (HELM-52)

mindburnlabs:
## Problem

`taintTrackingEnabled()` returned true only for an explicit `"1"`/`"true"`, so enforcement defaulted to **OFF**. Go's `&&` short-circuits at the call site, so with the variable unset `taintedEgressDenied` was never called and `TAINTED_DATA_EGRESS_DENY` never fired.

**An unconfigured kernel had no tainted-egress boundary at all** — while its own proof said otherwise. `proofs/GuardianPipeline.tla:52-53` states `TaintSafeEgress` with no flag or parameter, is listed in `Safety` (:172), enforced as an `INVARIANT` in `guardian.cfg:16`, and model-checked on every PR.

## Fix

| `HELM_TAINT_TRACKING` | Enforcement |
|---|---|
| unset, `1`, `true`, unrecognized | **enforced** |
| `0`, `false` | disabled |

An unrecognized value enforces so a typo cannot silently open a boundary. Disabling logs once at construction, naming the invariant it contradicts.

Renamed `taintTrackingEnabled` → `taintEgressEnforcementEnabled`: labelling is unconditional and unaffected; only the built-in deny was ever gated.

### The flip alone was not enough

Review found `allow_tainted_egress` was **absent from `IsReservedSecurityContextKey`**. Any transport that copies non-reserved caller arguments into the decision context and then marks it trusted — `pkg/mcp/server.go:87-94` does exactly this — let a caller pass the flag as a **tool argument** and self-approve its own tainted egress.

`fee619c9` closed the direct half by requiring a trusted context, but that marker means *"arrived via a transport"*, not *"this value was bound by the transport"*. Reserving the key makes such a transport reject it at the boundary.

**Without this, the doc change in this PR would have published a false security guarantee.**

## Scope of the proof claim

This **narrows** the code-vs-proof divergence; it does not close it. Three paths still ALLOW where the spec forbids — explicit opt-out, the transport-bound override (the spec models no override), and a request with no `destination` key. None are introduced here; each needs its own issue.

**No TLA+ change is required, and none should be made** — modelling the flag or the override would mean weakening `TaintSafeEgress`. Pipeline shape unchanged (6 gates).

## Tests

Four cases pin the default: unset enforces · unrecognized enforces · `"0"` disables · `"FALSE"` disables. The disabling cases assert `Equal(ReasonNoPolicy)` rather than `NotEqual`, which would have passed for any unrelated breakage. Plus reserved-key coverage (override still honoured when transport-bound, still ignored when not) and a case pinning that labelling runs with enforcement disabled.

## Gates

`make lint` · race on guardian/mcp/firewall/effects · `make verify-boundary` **1051 current, 0 stale, 0 errors** with the manifest regenerated in-commit.

## Two corrections to my own earlier claims, both caught in review

- **The bench figure was wrong.** "p99 26µs" was `ed25519_sign_only`, not the overhead benchmark. Actual overhead p99 ≈ **129µs** on this branch vs **122µs** on main — no regression here, but overhead p99 has exceeded the 75µs operating bar since June on both. Separate pre-existing issue.
- **`TestEvidencePackSingleSource` does not fail on pristine main.** I reported it as a pre-existing main-branch failure. It isn't — the local failure was a stray `.claude/worktrees/` copy of `evidence.go` that the test's `skipDirs` doesn't exclude. Test fragility, not a defect.

Rebased onto **v0.7.4** (`a079cdfc`), which shipped mid-review. The manifest conflict was resolved by regenerating, never by merging hashes.

Follow-ups filed by the review, not in scope here: snapshot the env at construction (it's read per-decision, so enforcement can change mid-process without appearing in the receipt); `destination`-key evasion; `TestEvidencePackSingleSource` walking `.claude`.

Refs: HELM-52, HELM-48

### DIFF
--- a/core/pkg/guardian/guardian.go
+++ b/core/pkg/guardian/guardian.go
@@ -252,6 +252,14 @@ func NewGuardian(signer crypto.Signer, ruleGraph *prg.Graph, reg *pkg_artifact.R
 		slog.Warn("[guardian] no authority clock injected; falling back to wall clock (KERNEL_TCB §3) — inject WithClock in production")
 		g.clock = wallClock{}
 	}
+
+	if !taintEgressEnforcementEnabled() {
+		// Logged once here rather than per-decision: the check sits on the hot
+		// path, but a disabled security boundary must never be silent. This
+		// state contradicts TaintSafeEgress in proofs/GuardianPipeline.tla.
+		slog.Warn("[guardian] tainted-egress enforcement DISABLED by HELM_TAINT_TRACKING; TAINTED_DATA_EGRESS_DENY will not fire (contradicts proofs/GuardianPipeline.tla TaintSafeEgress)")
+	}
+
 	g.boundaryChain = []BoundaryInterceptor{
 		g.zeroidInterceptor,
 		NewTemporalInterceptor(g),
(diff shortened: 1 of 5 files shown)


mindburnlabs:
Superseded by the DCO-signed replacement (same content; the `Signed-off-by` check rejected the unsigned commit and force-push is blocked in this workspace).

PR #639, opened 2026-07-21
ci: run the race gate and enforce reason-code reachability

mindburnlabs:
Two enforcement paths that existed and never ran. From the enforcement-integrity audit in `docs/ai/helm-verifiable-execution-research-review.md`.

## `go-race` was registered but unreachable

`go-race` is defined in `scripts/ci/quality-gates.json` with the correct command, 30-min timeout, and `core/**` path filter. It is assigned to exactly one profile — `merge` — and **no workflow invokes `make quality-merge`**:

| Profile | Workflow | Has `go-race` |
|---|---|---|
| `pr` | ci.yml:74 | no |
| `nightly` | nightly-quality.yml:44 | no |
| `release` | release.yml:84 | no |
| `merge` | **none** | **yes** |

So the race detector has never run in CI. Added as its own `kernel-race` job rather than a step on `kernel`, so it runs concurrently rather than serializing ~30 min onto the critical path.

**Verified:** `make test-race` passes clean on `main` today — 0 data races, 0 failures — so this lands green rather than red.

## Reason-code reachability (new)

48 of 96 declared `ReasonCode` constants are never emitted by non-test code. A code that is declared, enumerated, and given a conformance vector but never returned reads as coverage in review and produces nothing at runtime. `ERR_ASSUMPTION_STALE` is the canonical case: declared at `verdict.go:138`, documented in `reason-codes-v1.json:579`, vectored at `negative_vectors.go:188`, zero emit sites.

The gate requires each declared code to be emitted or recorded in `reason-codes-known-unreachable.txt`. It **also fails when an allowlisted code becomes reachable**, so the list burns down instead of rotting.

Matching is deliberately qualified — two packages shadow these names and must not count as emissions:
- `core/pkg/conform/reason_codes.go` reuses the Go identifiers
- `core/pkg/connectors/ton/acton/errors.go` reuses the wire values

An emission is `contracts.<Ident>` anywhere, or a bare `<Ident>` inside `core/pkg/contracts/` only. `ExpectedReasonCode:` lines are conformance expectations, not emissions.

### The allowlist records *why*, per group

The 24 `ReasonTON*` entries are **a duplicated registry, not unimplemented work** — `acton/errors.go` declares the same 24 wire values on its own type and the connector emits those. Remedy is collapsing the duplication, not adding emit sites.

Note `ReasonPlanTransactionConflict` is deliberately *not* allowlisted: it is emitted, at `bridge.go:371` — for nonce replay rather than a plan-transaction conflict. Reachability cannot catch a miscoded emission; that is tracked separately.

## Verification

- `make lint` — clean
- `make test-race` — clean (0 races)
- Gate green: `96 declared, 48 emitted, 48 allowlisted`
- Canary tested all three modes: new unreachable code fails; allowlisting it passes; a stale allowlist entry fails. `verdict.go` restored, no diff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,6 +139,8 @@ jobs:
         run: make lint
       - name: ProofGraph signature write guard
         run: bash scripts/ci/check_proofgraph_signature_writes.sh
+      - name: Reason-code reachability
+        run: python3 scripts/ci/check_reason_code_reachability.py
       - name: Build
         run: make build
       - name: Test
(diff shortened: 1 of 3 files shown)


mindburnlabs:
Superseded by #640 — same change, rebuilt on a fresh branch with a Signed-off-by trailer (DCO requires it on every commit, and force-push is blocked in this repo).

PR #641, opened 2026-07-21
feat: agent process ownership — worktree isolation, harness, fail-closed patch delivery (HELM-319)

mindburnlabs:
Closes part of [HELM-319](https://linear.app/mindburn/issue/HELM-319) · parent [HELM-76](https://linear.app/mindburn/issue/HELM-76) (Workstation Guard R6)

## What this is

HELM's native-client integration today installs HELM **into** a vendor coding agent — `setup_cmd.go:556` shells `claude mcp add … helm-ai-kernel mcp serve` and writes a `PreToolUse` hook. That governs only the calls the client chooses to route through it.

This PR adds the inverse: HELM **spawns** the vendor CLI as a child process it parents. A hook governs what an agent discloses; a parent governs what it can do.

Verified before starting: `grep exec.Command("claude"|"codex"|"cursor-agent"|"opencode")` across helm-ai-kernel, helm-ai-enterprise and svc-* returned **zero hits**. Nothing in the estate spawns a vendor coding CLI today.

Hooks are **not** replaced — they remain the only path for sessions HELM did not launch.

## Three packages

**`core/pkg/worktree`** — isolated git worktree at a recorded `BaseSHA`, plus a scoped `HOME` that is a **sibling** of the tree, never a child (vendor CLIs write credentials into `HOME`; inside the tree the `git add -A` in capture would pull them into a diff — proven behaviourally, not just by path arithmetic). Diff capture reads **raw bytes**: a line scanner splits on lone `\r` and rejoins with `\n`, corrupting CRLF at the point of capture. CRLF and binary round-trip are pinned.

**`core/pkg/harness`** — adapter contract + claude/codex. Three security properties, each **mutation-tested** to prove the tests aren't decorative:
- `killTree` signals the **negative** pid so grandchildren die with the leader (positive-pid mutation → 30s hang)
- SIGTERM escalates to SIGKILL on a timer (removing escalation → test fails)
- Env scrub is **cross-provider** (narrowing to `ANTHROPIC_` → `OPENAI_ORGANIZATION` reaches the child). Proxy/CA vars deliberately survive.

Read-only is **probed, not assumed** — if the CLI lacks the flags that make it real, the run is refused rather than run unenforced. `ObservedModel` is never backfilled from the requested model.

**`core/pkg/patchdelivery`** — fail-closed apply gate. The distinction it exists for:

| Verifier result | Meaning | Overridable |
| --- | --- | --- |
| `AppliedCleanly == false` | **Proven** conflict | **Never** |
| `AppliedCleanly == nil` | Verifier **errored** | Yes, accept-risk |
| `GatesPassed == nil` | No gates configured | n/a — never reported as passed |

**Overrides may bypass UNKNOWN, never PROVEN-FALSE.** Override is bound to `sha256(patch)`; one mutated byte invalidates it — `EffectPermit` semantics for a filesystem write.

## Naming

Two intended names were already taken by unrelated domains. Both distinctions are real and documented in the package comments:

| Intended | Taken by | Landed as |
| --- | --- | --- |
| `envelope` | **Autonomy Envelope** — signed runtime boundary *contract* | `worktree` |
| `delivery` | **Progressive release rollout** (MIN-720) | `patchdelivery` |

A run *binds* an Autonomy Envelope and *executes in* a Tree. `delivery` ships a release to an environment; `patchdelivery` ships a patch to a working tree.

## Verification

```
go build ./...                     ok
gofmt -l  {worktree,harness,patchdelivery}   (clean)
go vet    {worktree,harness,patchdelivery}   clean
go test   worktree        ok   12 tests
go test   harness         ok   31 tests
go test   patchdelivery   ok   17 new / 147 pkg total
go test -race             ok
```

Adjacent packages re-run and unaffected: `delivery`, `envelope`, `contracts`, `effects`.

## ⚠ Reviewer attention

1. **`MCPInjectionRequiresFullAccess` on both adapters is `UNVERIFIED`** — inherited empirical posture, marked as such in source, **pinned by no test**. Codex `true` (seatbelt cancels the server's call back to its host below full access), claude `false`. Must be re-proven against live CLIs before any claim depends on them. I corrected an inverted initial implementation but could not verify either value here.

2. **`ClientLoadObserved` does not exist.** Grepped `.go`/`.ts`/`.json`/`.md` estate-wide — **every hit is Markdown**. The docs describe "signed lifecycle receipts" carrying it, but no signed structure does, in any language. R6.1 must *introduce* it on `contracts.AgentRunReceipt` (protected path, needs kernel-guardian), not flip it. Not in this PR.

3. **Docs deliberately unregistered.** `docs/architecture/agent-process-ownership.md` is **not** in `public-docs.manifest.json` — that publishes to helm.docs.mindburn.org and this describes unshipped internals. May still need a ledger row in `Mindburn-Labs/docs` for the estate-wide docs-truth gate; flagging rather than guessing.

4. **Unix-only** — `syscall.Kill`/`Setpgid` with no build tags. Release binaries target linux/darwin.

## What this does NOT establish

No sterile native-client session proof. No receipt emission, ProofGraph node, or EvidencePack. No Guardian routing, no `EffectPermit` issuance, no signed verdict. Not wired to any CLI command or route.

**The estate's `ClientLoadObserved=false` boundary language stands unchanged.** No doc, issue, or public copy may claim HELM governs a native coding-agent session on the strength of this PR.

## Prior art

Boundary model follows [razzant/claudexor](https://github.com/razzant/claudexor) (MIT), reimplemented in Go. Attribution is a single line per package doc; no upstream identifiers or naming carried over.

One deliberate divergence: that project documents config-owned `protected_paths` as a user-facing approvals feature, but its producer is hardcoded empty (`orchestrator.ts:1608`), so the feature is inert. HELM does not port it — human approval belongs on `boundary/approvalceremony`, where approvals are *spendable* rather than merely checked, which composes with `EffectPermit` single-use/nonce semantics.


peycheff-com (approved):

mindburnlabs:
Closed as a preserved semantic predecessor of #772, now merged at d4b9a337. The historical recovery worktree and recovery/helm-pr-641-r61-envelope-wip-20260801 ref remain intentionally untouched because they contain dirty/untracked recovery work; this action does not delete or discard that material.

PR #650, opened 2026-07-22
fix(workstation): fail closed signer setup

SergeyAP:
## Summary

- Require `--signing-seed-file` for hook and setup signer resolution in `HELM_PRODUCTION`.
- Keep secure local signer provisioning limited to local and development paths.
- Propagate the approved signer-file path into installed hook, status, and removal commands.
- Add required `quantum_posture` annotations for the crypto-touching command surfaces.
- Cover production missing-key denial and explicit-key hook signing with ephemeral test keys.

## Security context

The existing v0.7.3 hardening removes the source-derived receipt seed, separates signature integrity from signer trust, and permanently rejects the retired deterministic signer as a trust anchor. This PR closes the remaining production setup/hook path that could otherwise generate a local key when a signer was absent.

Receipt canonicalization and signature preimages are unchanged.

## Validation

- `go test ./core/pkg/workstation ./core/pkg/runtimeadapters/mcp ./core/cmd/helm-ai-kernel -count=1`
- `make quantum-crypto-inventory`
- `make quality-pr`

## Required follow-up decisions

- Authoritative trust-store format and owner, including multi-key support.
- Production distribution mechanism for private signer and public trust anchor.
- Key rotation and migration plan; legacy receipts retain integrity verification but are intentionally untrusted.

Replaces #649.

Refs HELM-225

### DIFF
--- a/core/cmd/helm-ai-kernel/hook_cmd.go
+++ b/core/cmd/helm-ai-kernel/hook_cmd.go
@@ -1,3 +1,5 @@
+// quantum_posture: hooks configure only classical Ed25519 receipt signing;
+// they do not provide post-quantum or hybrid cryptographic protection.
 package main
 
 import (
@@ -20,9 +22,10 @@ var (
 )
 
 type hookOptions struct {
-	Client  string
-	DataDir string
-	JSON    bool
+	Client          string
+	DataDir         string
+	SigningSeedFile string
+	JSON            bool
 }
 
 type preToolPayload struct {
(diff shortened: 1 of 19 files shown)


SergeyAP:
Closing as an unassigned scope extension, not as delivery of HELM-225. The branch codex/helm-225-signer-hardening-signed and commit 9de245b1 are intentionally preserved for reuse. Follow-up: HELM-337 (Signer trust-store rotation and zero-seed hardening), which owns the required trust-store, key-distribution, rotation, migration, and ownership decisions before any selected code is merged.

PR #654, opened 2026-07-22
feat(hook): add --shell-mode=allowlist for default-deny shell classification

mindburnlabs:
Implements Workstream B of the "vs harness approvals" plan. Draft: this is a fail-closed classification change, so it needs `kernel-guardian` review before it leaves draft (CLAUDE.md).

## The defect

The PreToolUse hook governed `Bash` with a 12-substring denylist while treating `mcp__` tools **ten lines below** as default-deny with an allowlist. On a denylist miss, `classifyPreToolPayload` returned a zero classification and `hook_cmd.go` returned 0 with empty stdout — which under hook semantics means "no decision". The command ran with no verdict, no receipt, and no record that HELM saw it.

The reason nobody had closed it: `Observe.AllowedActions` holds abstract action IDs (`git.status`, `shell.read`) while the hook receives raw shell strings, and **nothing mapped one to the other**. The field was populated by the profile constructor at `importer.go:494` and read by nothing.

## What changed

- **`core/pkg/shellscan`** — `StaticallyAnalyzable` + `ClassifyReadOnly`. The metacharacter check runs first, so allowlist matching only ever sees a single simple command and the repo needs no shell lexer. It also rejects bare redirects and newlines, which the lifted `mcp` patterns do not cover: without that, `git show HEAD > ~/.ssh/authorized_keys` passes the metachar gate and then matches `git show` on the allowlist.
- The three shell regexes now live in `shellscan` and are imported by `core/pkg/mcp/argscan.go`, so the two surfaces cannot drift apart.
- **Honest reason codes** — an unanalyzable command denies with `SHELL_COMMAND_NOT_STATICALLY_ANALYZABLE` in the *signed receipt*, instead of falling through to the generic `OPERATE_PERMISSIONS_EMPTY`.
- **Policy file** — the hook reads `<data-dir>/policy/workstation.json` with fallback to the built-in default, and reports a profile-load failure separately from a signer failure rather than misattributing it as "local receipt signer is unavailable".
- **`setup` dedup fix** — hook identity is now a marker, not the exact command string. Exact matching meant a changed command line installed a *second* PreToolUse entry beside the old one with both firing, and uninstall then matched neither.

## Migration

`denylist` stays the default through 0.7.x, so no upgrade changes behavior. `0.8.0` flips the binary default to `allowlist` — that is what carries the change to installs whose baked command line was never rewritten. `--shell-mode=denylist` remains the escape hatch.

## Verification

- `make test-race` — clean, no data races
- `make crucible` — 12/12 use cases pass
- `make docs-coverage docs-truth` — 684 rows resolve
- `go vet ./core/...` — clean
- `FuzzClassifyReadOnly` — 2.36M execs, no failures. It caught a real flaw in the *test* on first run (`git stAtus rm -r`: a pathspec named `rm`, harmless), which is why the invariant now checks the executable position and table contents rather than substrings anywhere in the command.

## Deliberately not claimed

Pinned in `shellscan_gap_test.go` rather than papered over: classification bounds the *shape* of a command, not the behavior of the program it starts (`go test`, `make build`), and ambient config outside argv (`diff.external`, `GIT_EXTERNAL_DIFF`) is outside what this can see.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,12 @@ non-allowlisted HTTPS request. Prompt-only manipulation, model refusal, or an
 agent that never attempts the tool call needs model, app, and sandbox controls
 alongside HELM.
 
+Coverage differs by effect class. The v0.7 shell guard is a selected-effect
+guardrail: it recognizes a small set of literal destructive command forms and
+does not interpret shell expansion, aliases, `eval`, or wrappers. See
+[Workstation governance](docs/reference/workstation-governance.md) for the
+per-class boundary.
+
 ## One Example
 
 ```text
@@ -76,6 +82,7 @@ You get:  a signed receipt you can verify offline (integrity always;
 | Local AI-agent risk audit | [Agent Risk Scan](docs/reference/agent-risk-scan.md) |
 | CLI commands | [CLI reference](docs/reference/cli.md) |
 | Security model | [Execution security model](docs/EXECUTION_SECURITY_MODEL.md) |
+| Why not just an approval prompt | [Agent permission modes](docs/AGENT_PERMISSION_MODES.md) |
 | MCP tool quarantine | [MCP integration](docs/INTEGRATIONS/mcp.md) |
 | Client implementation handoff | [Implementation partner handoff](docs/guides/implementation-partner.md) |
 | Evidence verification | [Verification](docs/VERIFICATION.md) |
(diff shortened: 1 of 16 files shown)


peycheff-com (dismissed):

peycheff-com:
Closing as superseded; do not rebase or cherry-pick. Current main already has an AST-based shell scanner and signed allow/deny hook receipts, replacing this PRs regex scanner, hook routing, receipt, and deduplication. An opt-in strict shell policy remains a possible fresh narrow successor. The clean linked worktree and branch are intentionally retained pending separate cleanup proof.

PR #661, opened 2026-07-23
feat(kernel): list effect disposition candidates

peycheff-com:
## Summary
- add an internal workload-authenticated candidate listing route for effect dispositions
- reuse the existing verified effect reservation and disposition stores instead of introducing a new reconciliation state model
- return current reservation head data, current fence state, and next-chain fields needed for fail-closed reconciliation planning

## Validation
- cd core && GOWORK=off go test ./pkg/boundary/approvalceremony -count=1
- cd core && GOWORK=off go test ./cmd/helm-ai-kernel -run 'TestEffectDisposition|TestApprovalGrantConsumption|TestApprovalDispatchAdmission' -count=1
- git diff --check

## Boundaries
- internal_non_production route only; no public/operator route activation
- read-only source slice: no connector cancellation, compensation, reconciliation execution, or control-plane projection
- command recording still revalidates the exact fence and reservation head before accepting any disposition


peycheff-com:
Superseded by #662. This draft returned raw reservation/disposition payloads and permitted an unfenced listing, so it does not meet the required fail-closed reconciliation boundary. #662 replaces it with a FENCE-required, workload-scoped, typed `execution_authority: NONE` projection that exposes only minimal immutable command bindings.

PR #662, opened 2026-07-23
feat(kernel): project fenced reconciliation candidates

peycheff-com:
## Scope

Adds a narrow, internal Kernel projection for planning a later `RECONCILE_SOURCE` record after a durable FENCE. It returns a typed snapshot only after workload identity, FENCE scope, reservation authority, and existing receipt-chain checks pass.

## Safety properties

- Workload-JWT identity is the only scope source; query parameters are rejected.
- A current durable FENCE and receipt-integrity check are required.
- Only `STARTED` and fully bound `UNCERTAIN` reservations are included; `ADMITTED` is deliberately excluded.
- The response carries `execution_authority: NONE` and minimal immutable command bindings, never raw reservation/admission/release/grant payloads.
- A later record still rereads the FENCE, reservation head, and chain before accepting any command.

## Not included

No Data Plane relay, Control Plane outbox/projection, public/operator route, cancellation, compensation, connector action, deployment, or runtime proof.

## Validation

- focused contracts/boundary/HTTP tests
- `make verify-effect-disposition-vectors`
- `GOWORK=off go test -race ./pkg/boundary/approvalceremony`
- `GOWORK=off go test ./...`
- `make lint`
- `git diff --check`

The explicit real-Postgres proof was not run because `HELM_TEST_POSTGRES_URL` is unset; no fallback database was used.


peycheff-com:
**Cross-repo pairing note (2026-07-23, triage agent):** this PR pairs with **svc-helm-control-plane #228** (`feat(control-plane): observe FENCE-bound reconciliation candidates`).

The contract relationship, verified from both branches:

- **#662 (producer, Kernel):** defines the v1 `effect-reconciliation-candidates.v1` contract (`core/pkg/contracts/effect_reconciliation_candidates.go`) and projects FENCE-bound candidates via approval-consumption routes.
- **#228 (consumer, Control Plane):** accepts *only* that Kernel v1 contract, requires verified matching FENCE state plus `execution_authority: NONE` and the distinct `helm.effect.reconciliation.read` scope, and persists canonical snapshots in a default-off outbox/projection (migration 050). It creates no dispatch/cancel/compensation capability.

**Suggested merge order: #662 → #228.** Land the Kernel producer contract first so the Control Plane consumer merges against a tagged producer head; #228 is default-off and explicitly defers activation until the companion route and read scope are independently deployed and validated, so the gap between the two merges is safe. If #228 must land first for scheduling reasons, its default-off posture keeps that safe too — but producer-first keeps provenance cleaner.

Current state: this PR is MERGEABLE, 2 behind main, all checks green, still draft. #228 is now marked ready for review (0 behind main).

mindburnlabs:
Superseded by #762, merged as 1e63e482bf43e566f6db46c61b30ccb0ce24b383. The successor retains only the default-off, route-scoped, read-only reconciliation projection and adds exact resource binding; the older stale docs and unrelated reservation ListActive change were intentionally not carried into the release lane.

PR #688, opened 2026-07-25
feat: denials that carry what an agent needs to learn from them (HELM-373)

peycheff-com:
Producer half of a joint policy-reafference RFC with Mindot ([Will](https://github.com/mindot-ai/will)). Will's `.TODO/ENVELOPE_NARROWING.md` is already written against a HELM denial field that HELM does not emit. This adds it.

A denial currently tells an agent to stop and nothing else. The agent that exceeded a TTL ceiling by sixty days and the agent that touched a forbidden host get the same answer, and both retry at random.

## What lands

Two optional fields on denials, opt-in per policy profile, **absent rather than empty** when disabled so no stored receipt hash moves:

- **`finality`** — which of five kinds of "no" this was, derived from the rule that fired
- **`counterfactual`** — the nearest allowed envelope, for scalar bounds and required-capability names only

| finality | consumer behaviour |
| --- | --- |
| `class_forbidden` | erase stored bounds; stop probing |
| `ungranted` | needs an authority to say yes |
| `instance_parameter` | record the bound; retry within it |
| `instance_context` | touch nothing; the refusal was not about the action |
| `instance_membership` | that target is closed, others may work; the set is never described |

`ungranted` is terminal at the workstation boundary, which has no approver. At the kernel PDP the same finality rides `VERDICT_KIND_REQUIRE_APPROVAL` with the required signers attached. The verdict kind says what to do; the finality says what to learn.

## The disclosure boundary

Refusals that turn on **membership in a confidential set** — egress allowlist, workspace roots — disclose nothing beyond the reason code. An allowlist is a map of internal infrastructure, and "not that one, try these" would turn every refusal into a free probe of the estate. This is encoded as a per-code disclosure class whose zero value discloses nothing, so a code added without thought leaks nothing, and asserted as a property of the catalog rather than of the cases someone remembered to write. A disallowed **memory class** is not membership — the category comes from a fixed public vocabulary — so it stays `class_forbidden`; that distinction is the documented classification principle.

## Commits

| | |
| --- | --- |
| `b2f38d22` | workstation: classify all 12 deny reason codes, emit counterfactuals |
| `f530725c` | contract: additive fields on `DenyPayload` + `RequireApprovalPayload`, `DenialFinality` enum, `DenialCounterfactual` message |
| `7e99faa7` | conformance: `tests/conformance/denial-legibility` — frozen snapshot, golden, replay |
| `67be8bc5` | fixes from `kernel-guardian` and `conformance-engineer` review |
| `1e6e5bd1` | boundary manifest repin |
| `63a86927` | fifth finality value `instance_membership`; membership codes reclassified; `DenialCounterfactual` ↔ `CounterfactualReceipt` cross-pinned docs |
| `2e6b5cc3` | registry: per-code `finality` + `disclosure` attributes (57/96 classified), legibility invariants in `check_reason_codes.py`, vocabulary-parity conformance test |
| `54b5b5a4` | boundary manifest repin for the six changed files |

## Review findings, and what they caught

Two real defects in the first three commits, both fixed in `67be8bc5`:

**Finality was caller-steerable.** `normalizeEvents` lets the agent's own event log override the evaluated reason code, and the annotation consumed the overridden value. A producer could write one string into its own log and have the signed receipt tell the consumer `instance_context` — "this was not about the action, keep going" — about a boundary it had just been refused at. Finality now derives from the evaluated code. `TestDeclaredReasonCodeCannotSteerFinality` fails with the fix reverted; verified by mutation.

**The published JSON Schemas rejected the feature.** `denied_effect` and the workstation policy profile are both `additionalProperties: false`. Enabling the switches produced receipts every conforming consumer would reject, and this branch's own conformance snapshot failed its own contract. Both schemas gain the optional fields; existing documents validate unchanged.

Four narrower ones: `capability` echoed the raw producer-supplied `effect_type` for unrecognised types; an undeclared TTL was reported as "requested" after falling back to the profile default; the conformance pack bypassed the opt-in gate so its opt-out check was tautological; and the pack could be trimmed and its golden regenerated without any check noticing a whole finality class had lost coverage.

## The two contract decisions — resolved by the DRI (Ivan, 2026-07-28)

1. **Set-membership refusals get a fifth value, `instance_membership`** (`63a86927`). `class_forbidden`'s lesson — erase bounds, stop probing — was wrong for them: allowlisted hosts keep working. The classification principle: membership refuses a caller-chosen *target* against a set the policy keeps private; `class_forbidden` names a *category of action* from a fixed public vocabulary, which is why `MEMORY_CLASS_DISALLOWED` deliberately stays `class_forbidden`. Added now while it is additive; after any receipt ships it would be a v2 break. The RFC outline shared with Mindot is updated to five values.
2. **`DenialCounterfactual` keeps its name.** The wire field `counterfactual` is what Will's consumer draft is written against. The collision with `CounterfactualReceipt` is handled where it bites — both types now carry cross-referencing doc comments pinning the distinction: one rides enforced denials, the other records would-have verdicts and must never present as enforced.

## Registry attributes (previously blocked, now landed)

The domain-drift fix (HELM-374, #687) merged and is in this branch, so the registry work became actionable (`2e6b5cc3`): per-code `finality` on the 57 codes whose lesson is unambiguous — absent means unclassified, and an unclassified denial teaches nothing — and `disclosure` explicitly `none` on the membership codes only, absent elsewhere until a kernel-side counterfactual producer exists. `check_reason_codes.py` gains the pairing invariants the schema cannot express (`scalar_bound` ⇒ `instance_parameter`, `capability_name` ⇒ `ungranted`, forbidden/membership never disclose), and a conformance test pins the finality vocabulary across the Go constants, the receipt schema, and the registry schema.

## Not in scope

- **PDP wiring** is not actionable here. `verdict.proto` is schema-only in this repo; the producer is the Rust `crates/helm-policy-vm` behind a `cpi_native` cgo bridge, and that crate is not in this repository. Nothing here constructs a `DenyPayload`.

## Gates

`go test ./core/pkg/workstation/... ./core/pkg/contracts/...` · full `tests/conformance` suite (13 packages) incl. the new pack and parity test · `python3 scripts/ci/check_reason_codes.py` pass · `make verify-fixtures` exit 0 · `make proto-lint` exit 0 · `make proto-breaking` pass (WIRE + WIRE_JSON, v1 additive-only) · `make verify-boundary` 1056 current / 0 stale · `gofmt`, `go vet` clean.

`policy-schema` sits outside `PROTO_DIR`, so no SDK regeneration is involved.


peycheff-com:
## CI triage

**`Quality PR profile` and `kernel`** failed on boundary-manifest drift and are fixed in `1e6e5bd1`.

I did *not* use `tools/boundary/generate-manifest.sh`, which is what the error message tells you to run. On this machine it emits a manifest roughly a third the size of the committed one — the diff was 32 insertions against 764 deletions, i.e. it would have silently dropped ~760 other people's hashes while looking like a clean regeneration. Instead the four stale digests were recomputed in place; `verify-boundary` then reports **1056 current, 0 stale**.

Worth someone's attention separately: a repair tool that quietly truncates the artifact it repairs, on a host that isn't CI, is a trap the next person will also walk into. (The commit message cites "424 of 1056" for the generator's coverage — that figure came from a truncated verify run; the accurate statement is that the generated manifest is ~330 entries against 1056.)

**`deployment-smoke`** was a Docker Hub reachability flake, not a regression:

```
#2 ERROR: failed to do request: Head "https://registry-1.docker.io/v2/library/golang/manifests/sha256:5696...":
   dial tcp 98.91.123.139:443: i/o timeout
```

It passed on the previous head and the job only runs `docker-smoke compose-smoke helm-chart-smoke`, none of which reads the boundary manifest. Re-run.

**`Deterministic repository gates` and `HELM Autonomous Release Permit`** are the merge-authority interlock, not code defects. They fail at `Verify exact merge-parent binding` and `Download permit context` (the second cascades from the first). Neither job is defined in this repository's `.github/` — they arrive as an org-level required workflow. Per `CLAUDE.md`, covered merges need the distinct-provider 2-of-2 permit and the exact-head approval-only App interlock, so these cannot go green from inside the PR: they need two approvals from distinct providers and a permit bound to the exact head SHA.

## Local gates on `1e6e5bd1`

`go test ./core/pkg/workstation/... ./core/pkg/contracts/...` · full `tests/conformance` suite · `make verify-fixtures` · `make proto-lint` · `make proto-breaking` · `make verify-boundary` · `gofmt` · `go vet` — all clean.

One caveat on that conformance line, since this PR adds a pack: **`tests/conformance` is not run by any CI workflow.** It is a separate Go module, `make test` does not enter it, and `codeql.yml` only compiles it for static analysis. So the new `denial-legibility` pack — like `profile-v1` and the twelve others — passes only when someone runs it by hand, while `tests/conformance/README.md` grants those same checks the authority to back public conformance claims. Tracked separately rather than bundled here.

mindburnlabs (dismissed):

PR #673, opened 2026-07-23
feat(shellscan): AST-based shell pre-flight classifier for the pre-tool hook

peycheff-com:
## Summary

Replaces the evadable substring needle list in `core/cmd/helm-ai-kernel/hook_cmd.go` with a structural bash AST classifier (new `core/pkg/shellscan`) as **advisory input** to the existing signed decision path. Verdict authority is unchanged — `workstation.Decide` still produces the permit/receipt verdict, fail-closed; the classifier only decides whether a shell command reaches that path.

Evidence base: `research/opencode-study/03-core-tools.md` (opencode's tree-sitter shell scan + arity dictionary) and `research/opencode-study/25-helm-map-agent-side.md` (A1: needle list trivially evadable — `rm -r -f`, `xargs rm`, `find -delete`, variable indirection; A2: arity table; guidance to keep the needle list as a fallback layer).

## What shipped

- **AST parse of every command** via `mvdan.cc/sh/v3` (pure Go, zero cgo; chosen over tree-sitter Go bindings to keep the TCB small and the build cgo-free; only new transitive deps are `golang.org/x/sys`/`golang.org/x/term`, already in the module graph). Walks all command nodes including pipelines, subshells, and command substitutions.
- **Arity-aware command prefixes** (`git checkout`, `npm run dev`, `docker compose up`) via a Go port of opencode's arity dictionary — groundwork for permit scoping (A2).
- **Evasion closure**: split/reordered/long flags (`rm -r -f`, `rm --recursive --force`, flags after operands), `sudo`/`doas`/`env`/`nice`/`nohup`/`stdbuf`/`setsid`/`busybox`/`xargs` wrapper unwrapping, recursive `sh -c`/`bash -c`/`eval` payload classification (depth-capped), base64/xxd/openssl decode-to-shell pipelines, path obfuscation (`/bin/./rm`), `find -delete` / `find -exec rm -rf`, and write-redirects to sensitive targets (`>> .env`) that bypassed Write-tool path checks.
- **Fail-closed on the opaque**: unparseable input, dynamic command words (`$(echo rm) -rf /`), dynamic flags/operands on destructive families, and dynamic `eval`/`sh -c` payloads all route to the decision path.
- **Zero regression by construction**: the legacy needle list is kept verbatim inside shellscan as a fallback OR-layer; AST detection is strictly additive.

## Test evidence

[...]

PR #701, opened 2026-07-26
feat(releasepermit): defer P2 instead of blocking on it (schema v4)

peycheff-com:
## Why

The distinct-provider release permit counts P0–P2 as blocking. On any substantial change that makes the gate unsatisfiable by iteration: each round of fixes closes the reported findings and the next review returns a fresh crop, so the blocking set never drains.

Evidence from the current backlog (36 permits scanned, findings pulled from the review artifacts):

| | anthropic/claude-fable-5 | openai/gpt-5.6-sol |
|---|---|---|
| deny rate | 14% | 64% |
| blocking findings raised | 7 | 39 |
| severities used | P2, P3 only | P1, P2 |

19 of 24 denials were single-handed from one lane. Three kernel PRs carry `permit round 9` in their commit history; a tenth round (#671, #672, #673 — fixes pushed 2026-07-26) produced three new P1s and zero repeats of the findings it closed.

## What changes

- **P0/P1 keep denying.** Catastrophic/irreversible and exploitable security, authorization, data-integrity or release-gate failures are unchanged.
- **P2 becomes deferred.** Counted onto the permit as `ReviewSummary.DeferredFindings` so the defect stays attached to the record and is owed a follow-up — not a reason to hold a release.
- **A DENY verdict still denies.** `REVIEW_DENIED` is independent of severity, so the verdict stays the reviewer's call. This only stops the reducer from manufacturing a denial out of a P2 on an otherwise-ALLOW review.
- **Schema v3 → v4.** `deferred_findings` changes the permit's canonical form and therefore every `permit_id`; the golden vector is repinned. The strict decoder takes the new key explicitly — unknown keys are still rejected.

## Sequencing

This is one half of the change. The reviewer rubric in `Mindburn-Labs/.github` (`scripts/autonomous_release_permit.py`) still says *"ALLOW only when you found no P0, P1, or P2 issue"* and must move to P0/P1 in the same authority generation. Until both land, the verdict and the reducer disagree about P2.

Follow-up order: merge here → bump the verifier `ref:` in `.github/workflows/ci.yml` alongside the rubric change → advance the ruleset's pinned workflow SHA.

## Not in scope

The quorum stays 2-of-2 distinct-provider. The calibration gap between the lanes is a separate question and is being tracked in Linear rather than fixed by widening the quorum.


peycheff-com:
Converting to draft. Both lanes denied at **P1**, and the second finding is correct on the merits:

> **GATE_POLICY_SELF_MODIFICATION** — *"The change relaxes the release-authority reducer's own admission policy from within a target-repo PR... Weakening the quorum gate should be ratified through the authority-generation promotion path, not merged under the gate it loosens."*

That is the designed path, documented on `AuthorityParent` in `types.go`: *"The previous generation must review and ratify the complete next generation before an external promotion broker advances the ruleset pin."* Generation 2 cannot ratify generation 3 through an ordinary pull request, and this PR tried to.

Per `Mindburn-Labs/.github` ADR `2026-07-18-phase1-ci-authority` item 3, the least-privilege GitHub App that would act as that promotion broker has not been created — *"An organization owner must create or nominate that identity."* So the promotion path is specified but not yet operable, and this change has nowhere legitimate to land until it is.

Two further findings worth carrying into whatever does land:

- **PERMIT_SCHEMA_SKEW (P2)** — a v3→v4 permit shape needs a stated compatibility story for verifiers pinned to older generations, which this patch does not give.
- **STRICT_SHAPE_TEST_GAP (P3)** — no negative test proving an old-shape permit is rejected.

The substance stands and the evidence is in the PR body: across 36 evaluated permits the blocking set does not drain under iteration. Leaving this branch as the prepared generation-3 candidate rather than closing it. The paired reviewer-rubric change is in `Mindburn-Labs/.github` on `permit-p2-advisory-rubric`.

peycheff-com:
Tracked as [HELM-392](https://linear.app/mindburn/issue/HELM-392) — the measurement behind this change, the promotion-path blocker, and the two carried-forward findings (`PERMIT_SCHEMA_SKEW`, `STRICT_SHAPE_TEST_GAP`) are all recorded there. This branch stays as the prepared generation-3 candidate.

peycheff-com:
Closing after a fresh ruleset audit. The active main-protection ruleset does not list the HELM Autonomous Release Permit as a required GitHub status check, so this draft was based on an incorrect premise. We will keep the existing internal permit contract and fix product PR findings instead of weakening P2 handling.

PR #709, opened 2026-07-28
feat(guardian): state which gates produced each verdict (HELM-407)

peycheff-com:
## What

Every Guardian verdict now carries `GateRosterHash` — the digest of the gate set that produced it.

`Guardian` skips an uninjected gate rather than refusing it (every call site is `if g.<field> != nil`). Two kernels can therefore return an identical `ALLOW` from entirely different enforcement, and the decision, the receipt, and the EvidencePack all look the same. `GateRoster()` and `GateRoster.Hash()` already existed in `core/pkg/guardian/roster.go` for exactly this purpose — the doc comment there prescribes this step verbatim ("binding the roster hash into the decision record ... follow") — and had zero callers.

## Changes

- `contracts.DecisionRecord.GateRosterHash` — new field, following the `CorrelationID` precedent of a record field carried outside the signature.
- `NewGuardian` digests the roster once, after all options are applied, and logs it: active set, inactive set, active vs declared count, hash.
- `signDecisionWithGraph` binds the digest before any signing path. It is the single funnel all six `DecisionRecord` construction sites reach, so a seventh cannot miss it.
- `TestModelGateSetMatchesDeclaredGates` pins `proofs/guardian.cfg` against the declared `GateID`s in both directions.

## Why the spec test

`proofs/GuardianPipeline.tla:42` proves `AllowRequiresUnanimity` over the gate set in `proofs/guardian.cfg:7` and has no state for an absent gate. Nothing checked that those six model names still correspond to real gates, so a rename or a spec edit would silently narrow what TLC verifies on every PR without failing anything. Now it fails.

## Deliberately not in this PR

**The digest is outside the decision signing preimage.** `crypto.CanonicalizeDecision` is a positional format string; extending it is a preimage-version migration that belongs with `crypto.ReceiptPreimageV5` (`STATUS: NOT YET ACTIVE`, blocked on receipt-store columns). Until then the digest is tamper-evident via the receipt envelope chain hash, not signature-covered. Worth a follow-up issue.

**No proto change.** `core/pkg/contracts/decision.proto` already lacks `requirement_set_hash` and `policy_epoch`; this follows that precedent rather than opening a wire-contract change. If the digest should reach gRPC consumers, that is its own PR with `buf lint` + `go-apidiff`.

**This reports the gap, it does not close it.** The intersection of the six formally verified gates and the gates production injects (`WithPolicySnapshots`, `WithScopedStopReader`, optional `WithWarmLeaseManager`) is currently empty. This makes that machine-readable per decision; reconnecting gates is separate work.

## Validation

- `go build ./core/...`
- `go test ./core/...` — full suite green
- `make verify-fixtures` — all reference packs pass, including independent Python parity and negative mutations. Fixtures are unchanged, which is the evidence that the preimage did not move.
- `make lint` — `go vet`, gofmt, and docs coverage/truth checks green

Linear: [HELM-407](https://linear.app/mindburn/issue/HELM-407)

### DIFF
--- a/core/pkg/contracts/decision.go
+++ b/core/pkg/contracts/decision.go
@@ -57,10 +57,19 @@ type DecisionRecord struct {
 	SessionCentroidHash    string  `json:"session_centroid_hash,omitempty"`
 	RiskAccumulationWindow int     `json:"risk_accumulation_window,omitempty"`
 	// RequirementSetHash links this decision to the specific Proof Requirement Graph rules satisfied.
-	RequirementSetHash string    `json:"requirement_set_hash,omitempty"`
-	Signature          string    `json:"signature"`
-	SignatureType      string    `json:"signature_type"`
-	Timestamp          time.Time `json:"timestamp"`
+	RequirementSetHash string `json:"requirement_set_hash,omitempty"`
+	// GateRosterHash digests the Guardian gate roster (guardian.GateRoster)
+	// that produced this verdict, so evidence states which gates ran instead
+	// of leaving that to code review. An uninjected gate is skipped rather
+	// than refused, so two kernels can return the same verdict from different
+	// enforcement: without this the difference is invisible downstream.
+	// NOTE: outside the decision signature — folding it into the signing
+	// preimage is a preimage-version migration (crypto.ReceiptPreimageV5).
+	// It is still tamper-evident via the receipt envelope chain hash.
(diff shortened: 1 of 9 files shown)


peycheff-com:
Superseded by #710. This branch was cut from a local `main` that carried an unmerged commit (a85ca4e4, launchpad F-21), so the PR picked up five unrelated files. #710 is cut from `origin/main` and contains only the four intended files.

PR #713, opened 2026-07-28
fix(boundary): police forwarded egress, not just locally-generated

peycheff-com:
The compiled Boundary Enforcement Profile carried a single nftables base chain on the output hook. That hook only sees locally-generated traffic, so a workload in its own network namespace — any container on a bridge — egressed as forwarded traffic and was never policed, while the artifact still reported default-drop egress.

Measured on Linux 6.12 with a nested netns behind a masquerading veth:

| ruleset | host process | workload in its own netns |
|---|---|---|
| output chain only | BLOCKED | REACHED |
| output + forward | BLOCKED | BLOCKED |

A native nftables forward-hook drop also wins against an explicit accepting iptables-nft FORWARD rule, so this stays authoritative alongside a container runtime's own chains.

The allowlist is emitted on both hooks so one policy input cannot mean two different things depending on the path a workload takes. Loopback is exempted only on output — it is never forwarded.

Scoping limits are now recorded in the deployment doc rather than assumed: the forward chain matches destinations, not workload identity, and the kernel rejects `socket cgroupv2` on the forward hook, so per-workload scoping remains systemd-unit-only.

Reference pack regenerated; the independent Python verifier passes. Local gates green: `go test ./pkg/boundary/...`, `go vet`, `gofmt`, `tools/verify-boundary.sh` (1060 current / 0 stale), `make docs-truth` (683 rows).

Closes HELM-412.


peycheff-com:
Superseded by #714. This branch was cut from a local main that carried an unpushed commit (a85ca4e4, launchpad F-21), so the PR bundled unrelated work and conflicted with main. #714 is origin/main plus the single boundary commit, with all gates re-run on that base.

The branch is left in place rather than deleted — a85ca4e4 is not merged anywhere and should not be lost.

PR #715, opened 2026-07-29
fix(workstation): close destructive-command gaps and let scan see real machines

peycheff-com:
Closes three workstation-governance gaps found by replaying real agent behavior against the shipped PreToolUse hook and risk scanner.

**Destructive command coverage.** The hook now recognizes force-push forms, `drop database`, `dropdb`, `terraform destroy`, `terraform apply -destroy`, `aws s3 rm`, and namespace deletion. Matching covers compound commands, path-qualified executables such as `/usr/bin/git`, and fails closed on unknown `rm` options.

**Build-artifact cleanup without a broad bypass.** Routine recursive cleanup remains allowed for narrowly enumerated regenerable paths such as `node_modules`, `.next`, and `dist` only when the payload CWD resolves below the current user home inside a Git worktree. Absolute, parent-relative, variable, glob, wrapper, cwd-changing, mixed, symlink, and unrecognized operands are denied. Generic artifact names from `/srv/production` therefore fail closed, while the same cleanup in a real local source checkout remains silent.

**Risk scan sees real machines.** `scan` now reads the real Claude, Codex, desktop, and enabled-plugin configuration surfaces by default, with explicit user/project precedence and RFC3339-aware latest-installed-plugin selection. Missing optional local plugin state is treated as not installed; present malformed recognized configuration, including nested non-object `mcpServers`, fails closed. `--no-user-config` provides deterministic project-only scans. On the reference machine the command reports 23 MCP servers across 8 static config files instead of 0.

Validation includes focused adversarial regressions, `go test ./...`, documentation truth, boundary verification, real-machine scan smoke, and runtime hook proof: source-worktree cleanup emits no decision, while `/srv/production` cleanup emits a signed DENY receipt that verifies with `integrity:true` and `trusted:true`. Release permit is reissued for every fresh head/base pair.

Closes HELM-416.


peycheff-com:
Superseded by #751, rebuilt from exact current main. The successor uses the AST path only, covers the missing destructive command families, limits real-machine config discovery to named config locations with explicit project-over-user precedence, and includes fresh focused/race/boundary/docs validation. The original #715 branch and dirty recovery worktree are retained as recovery evidence.

PR #740, opened 2026-08-02
fix(crypto): keep the decision explanation attested under preimage V2

peycheff-com:
Follow-up to #736, which landed the rest of HELM-303. One gap survived it.

## The gap

`decision_record.v2` promoted `ReasonCode` into the signed payload and dropped free-text `Reason` from it. For a DENY carrying a registry code that is a net improvement. Everywhere else it removed attestation the legacy preimage had.

`ReasonCode` is empty **by contract** on ALLOW — [`contracts/verdict.go`](https://github.com/Mindburn-Labs/helm-ai-kernel/blob/main/core/pkg/contracts/verdict.go) defines the registry for DENY/ESCALATE, and `governance.Guardian.Authorize` clears it explicitly on the allow path. So a V2-signed ALLOW authenticates no explanation at all.

Reproduced against `main` before this change:

```
signature_version="decision_record.v2" reason_code=""
after rewriting Reason: verifies=true
```

The stated reason of an allowed action — `risk_score 12 < threshold 80 for action "deploy"` — can be rewritten to `operator override, no policy consulted` and the signature still verifies.

## The change

`Reason` is bound as `reason_hash`, a SHA-256 digest, rather than verbatim. Prose stays out of the preimage — the telemetry contract prohibits exporting it — while any edit to the emitted explanation breaks the signature, on every verdict.

The new field sits in lexicographic position, so the hand-rolled encoder still matches `canonicalize.JCS` byte for byte and `TestDecisionV2PreimageMatchesGenericJCS` keeps checking that. Guardian's allocation budget test still passes.

## Tests

Three existing tests asserted the gap as intended behaviour and are inverted. The DRIFT-7 tamper table gains a `Reason` case. `TestCanonicalizeDecisionV2_AllowExplanationIsAttested` pins the allow path specifically, since that is the one the code contract makes unavoidable rather than merely likely.

## Compatibility

Preimage-affecting. Records already signed under `decision_record.v2` will not verify against the new derivation. V2 shipped in #736 earlier today, so the exposure window is hours; no `v0.8.0` release carries it yet. Flagging explicitly rather than assuming — if any V2-signed decision has been persisted somewhere durable, this needs a version bump to `decision_record.v3` instead.


peycheff-com:
## Compatibility question in the description — resolved, no version bump needed

I flagged that this changes a preimage and asked whether durable `decision_record.v2` records exist. Checked rather than assumed:

- **Never released.** Latest tag is `v0.7.5` (2026-07-23). `git tag --contains 4fe7bf934` is empty — no tag carries #736, which introduced V2 at 19:04 UTC today.
- **The kernel does not persist decision records.** Only receipts are stored. Every `signature_version` column in `core/pkg/store` belongs to a receipt, and the only file under `store/` that mentions `DecisionRecord` is a test.
- **This PR does not touch the receipt preimage.** The diff is confined to `decisionV2SigningEnvelope` / `CanonicalizeDecisionV2`. `receipt.v5` — the version that *is* durably persisted, in the `receipt_envelope` column — is untouched.
- **No sibling service references it.** `svc-helm-control-plane`, `svc-helm-data-plane` and `platform-agent-substrate` contain no `decision_record.v2` / `DecisionRecordSignatureV2` reference.
- **`helm-ai-enterprise` carries its own vendored crypto**, pinned at kernel lock `a85ca4e4`, so it is unaffected until a deliberate sync.

So no durable V2-signed decision can exist, and amending `decision_record.v2` in place is correct rather than minting a `v3`. The exposure window was roughly four hours of unreleased `main`.

PR #799, opened 2026-08-06
feat(effects): sign the EffectPermit on issuance and verify it fail-closed

mindburnlabs:
The permit is the pre-action authorization artifact the G-Std submission is about, and nothing signed it. `effects.EffectPermit` carried a `Signature` field that no issuance path populated (`git grep "CanonicalizePermit\|SignPermit\|PermitSigningPayload"` returned nothing), so any process able to hand a connector a permit could mint its own authority, and a holder could widen scope or push out expiry without invalidating anything.

## permit.v1
A JCS-canonicalized narrow durable envelope, mirroring the receipt v5 design in `canonical.go`. It binds every field that decides what the permit authorizes: `permit_id`, `intent_hash`, `verdict_hash`, `policy_hash`, `effect_type`, `connector_id`, the **whole** scope (`allowed_action`, `allowed_params`, `deny_patterns`), `resource_ref`, `expires_at`, `single_use`, `nonce`, `issued_at`, `issuer_id`, `evidence_bindings`.

Carve-outs, each documented in the envelope's doc comment:
- `Signature` — cannot be an input to the value it authenticates.
- `PlanHash` — advisory planning provenance no enforcement path reads; if a connector ever gates on it, it moves into permit.v2 rather than being trusted unsigned.

Nothing else is excluded. Unlike a receipt there is no post-signing anchoring step, so no field legitimately changes after the signature exists.

Both receipt defects are tested against, not just avoided:
- **F-05** (v4 bound 8 of ~80 fields): `TestPermitSignatureBindsEveryAuthorizingField` mutates all 17 bound fields and requires each to break the signature.
- **F-06** (bare `:` join, no escaping, two records one preimage): `TestPermitPreimageResistsFieldBoundaryConfusion` moves a colon across a field boundary and requires distinct preimages.

`PermitSigningPayload` is the single stamped entry point. `PermitVerifyPayload` dispatches on the declared version and rejects unknown **and missing** versions — an unversioned permit predates signing entirely, so it is unsigned or forged.

## Consumption gate
`VerifyPermitForUse` denies on: unsigned, missing version, tampered field, foreign key, expired, replayed single-use nonce, wrong connector, out-of-scope action, out-of-scope parameter. Any error denies. Table-driven test covers all nine.

## Issuance
`GovernedBridge.mintPermit` signs through the existing `crypto` plumbing (`NewEd25519SignerFromSeed` over the bridge's configured seed — no new key handling) and **refuses to issue at all** when no seed is configured.

## Key-reachability limitation — stated, not papered over
In the workstation deployment the bridge's signing seed is a 0600 file in the kernel data directory on the same host and under the same user account as the agent whose actions are being authorized. A process running as that user can read the seed and mint permits that verify. This change makes permits unforgeable by a *remote* or *differently-privileged* party and makes tampering detectable, but it does **not** by itself make the key unreachable by a local agent. Closing that requires the kernel to hold its key under a separate uid or in an external signer; filed as follow-up work rather than claimed here.

## Gates
`go test ./pkg/crypto/ ./pkg/effects/ ./pkg/runtimeadapters/...` green; `make quality-pr` green; `make verify-boundary` green (manifest rows added in place for the three new files, no wholesale regeneration); quantum-crypto-inventory passed.


mindburnlabs:
Closing as a duplicate of #798, which is merged (`cd371d8ea`).

Both PRs implemented EffectPermit signing independently and neither referenced the other; they also both touched `core/pkg/runtimeadapters/mcp/bridge.go` and `tools/boundary/protected.manifest`, so landing both would have conflicted.

#798 won on two counts:

- **It carries the consumption side.** Signing alone does not close the gap — the permit has to be *verified* before an effect executes. #798 adds `verifyPermitSignature` in the connector path with tests for tampered, unsigned and foreign-signer permits (`TestExecuteRefusesTamperedPermitWithoutAttemptingEffect`). This PR signs but never verifies at the point of use.
- **It passes `quantum-crypto-inventory`.** This PR fails that gate; #798 carries the `quantum_posture` annotation the gate requires.

The canonicalization strengths of this PR are not lost — #798 arrived at the same design independently and, on inspection, has all of it: an explicit `effect_permit.v1` version constant inside the signed envelope for domain separation, a `effectPermitScopeV1` view without `omitempty` so a nil slice and an empty slice canonicalize identically instead of `null` vs `[]`, UTC normalization on both timestamps, and full field coverage with a comment stating that any new field must be added under a new version constant. It also documents a real caveat this PR did not surface: `evidence_bindings` is covered by the signature but has no counterpart in the proto contract, so v1 signatures are in-process only until the wire contract gains the field.

No work is being discarded — the two implementations converged. Thanks; the duplication was an orchestration fault, not a fault in this PR.

PR #803, opened 2026-08-06
fix(effects): give evidence_bindings a wire home so permit signatures cross a hop

mindburnlabs:
Closes **B2** of the standards-readiness audit.

`effect_permit.v1` signs `evidence_bindings`; `message EffectPermit` in `protocols/proto/helm/effects/v1/effects.proto` had no such field. Any proto or schema-validated JSON hop dropped a **signed** field, so a permit signed in one process could not verify in another — v1 signatures were in-process only, as `effect_permit.go` itself noted. Nothing caught it because the signed envelope and the wire contract live in different modules and no test compared them.

Field 16 is a backward-compatible add (older readers ignore it; newer readers reconstruct the exact signed preimage after a hop). Go bindings regenerated via `make codegen-go`.

## The test that would have caught it
- `TestSignedPermitFieldsExistInTheWireContract` reads the `.proto` and requires every field the signed envelope binds to exist in the wire contract. `core` cannot import the generated SDK package (separate module, and the dependency would point the wrong way), so it compares the two artifacts **as data**. Verified it bites: removing the new proto field fails it with `effect_permit.v1 signs "evidence_bindings" but message EffectPermit has no such field`.
- `TestDroppingASignedFieldInTransitBreaksVerification` strips the field in transit and requires verification to fail, so the parity check protects something real.

## Gates
`go test ./pkg/crypto/` green, `make proto-lint` green, `make verify-boundary` green (two digests repinned in place — the proto and the new test — no wholesale regeneration).

Audit note: this leaves the permit's **signed** field set portable. The receipt's normative-serialization question (B1) and the published vector corpus (B3) are separate follow-ups.


peycheff-com:
Resolved the four `github-code-quality` threads on `sdk/java/.../Effects.java` without changing the file.

That file is protoc output — its header reads `Generated by the protocol buffer compiler. DO NOT EDIT!`. Adding the suggested `@java.lang.Override` annotations by hand would be undone by the next `buf generate`, and would then fail the `SDK regenerate-and-diff gate`, which asserts the checked-in bindings are byte-identical to a fresh generation. The annotations are also emitted by protoc itself as a function of the protobuf-java version, not by anything in this change.

The diff here adds one field to the permit wire contract; the Java delta is entirely regenerated output.

PR #830, opened 2026-08-09
Trace the daemon serve path: a span, a route and a flush (HELM-495)

SergeyAP:
> **Security decision recorded.** The current shared listener treats every inbound `traceparent` as untrusted: the kernel starts a local root span and keeps the remote context only as a link. That fail-closed posture is approved for this source change. HELM-515 remains Active for the separate authenticated mesh edge needed before trusted ext_authz parenting can be claimed. Exact-head CI and live QA/store correlation remain independent gates.

Closes nothing on its own. Delivers the code half of [HELM-495](https://linear.app/mindburn/issue/HELM-495).

## The defect

The kernel produced **zero application spans on every environment**, with a correct OTLP endpoint, a provider that initialised cleanly, sampling at 1.0 and no export error anywhere.

The cause was not the exporter and not the transport: **no span was ever started.** There are two HTTP servers in this repo, and the instrumented one is never constructed —

```
$ git grep -n "NewServer(" -- 'core/**/*.go' | grep -v _test.go
core/pkg/api/server.go:195:func NewServer(cfg ServerConfig) *Server {
```

one definition, zero non-test call sites. HELM-333 attached `tracing.WrapEdgeHandler` to that dead constructor, while the daemon builds its own `http.Server` with four middlewares, none of them tracing.

The proof came from the collector side rather than from the code: `telemetry-gateway` had **no span counters at all** — not zero, *absent*. Collector counters are created on first use, so their absence means the receiver had never seen a span. A `telemetrygen` Job in the same namespace landed spans normally, which cleared the door, the gateway, WireGuard and ClickHouse in one step.

## What this changes

**The wrapper goes on the daemon's own chain, below `RequestIDMiddleware`.** That position is load-bearing, not incidental: `RequestIDMiddleware` passes a *copy* of the request, and `net/http.ServeMux` stamps `Pattern` onto the request it is given, so a wrapper above it never sees the route. Every layer below must keep forwarding the request `otelhttp` handed it; `rateLimiter.Middleware` does, and the layer-order guard says so in its own failure message.

**Route identity lives inside `WrapEdgeHandler`**, because two of its three surfaces cannot be reached from the call site: `otelhttp` re-applies its span-name formatter *after* `next.ServeHTTP`, overwriting any `SetName`, and the metric attribute is an option rather than a middleware. One normalizer — a deliberate mirror of the library's own `httpRoute()` — feeds the span name, the span attribute and the duration histogram, so the three cannot disagree.

Two things that look like over-engineering and are not:

- **Both spellings of a root catch-all are refused**, `/` and `/{rest...}`. A route that matches every path names the edge *worse* than the operation constant it replaces. The proxy edge registers `mux.HandleFunc("/", …)` as its terminal route, so without this the whole governed-LLM surface reports as `POST /` — and rewriting that route in the modern wildcard form, which reads like a harmless modernisation, would have reintroduced it. `/{$}` is deliberately kept: it means the root document and nothing else.
- **The method is folded onto the known verb set.** Bare-path patterns match *any* method, so without the fold a caller mints a distinct span name per request just by varying the request-line token.

**The otelhttp server metrics are bounded by an explicit attribute allowlist.** This one is a blocker found in pre-commit review, not part of the original scope: the default set carries `server.address` and `server.port` **verbatim from the client-controlled `Host` header**, so 64 requests differing only in that header produced 64 permanent series on each of three instruments — unbounded heap growth reachable by anyone who can open a socket to the API port.

An allowlist rather than a two-key denylist, deliberately: the metric attribute set is a subset of the *span* set, which already carries `user_agent.original`, `client.address` and `url.path`. The two have moved together before, and a denylist would walk a new key straight through on the next dependency bump.

**Three adjacent defects**, each of which would have left the fix half-delivered:

| | before | after |
| -- | -- | -- |
| `service.name` | pinned to `helm-sovereign-os`, `OTEL_SERVICE_NAME` ignored | read from the env, settable through the chart's `telemetry` block |
| SIGTERM flush | provider never shut down; the flush that was added shared a 10 s budget with the drain it waits behind, and `Provider.Shutdown` returned `nil` unconditionally so the failure was silent | own budget created *after* the drain returns, split so the trace half uses whatever the metric half does not need; `Shutdown` reports the first real error |
| request-path logging | no context passed, so the trace-aware slog handler had nothing to stamp | 61 sites on the `*Context` variants with the handler's own request |

**Four comments asserted a wrong mechanism as fact** and are corrected against the code. `otelhttp` does not capture its tracer at construction time — it resolves per request at v0.65.0. The SSE route cited as the reason for a separate flush budget does not exist under that name. And the metric attribute has *two* post-routing hooks, not one; the `Labeler` is the other, and it is the right hook for an attribute only one handler knows. Wrong-fact comments have cost this change a day twice, and one of these had already been "corrected" once into a different false absolute of the same shape.

## How it is verified

`go build ./...`, `go vet ./...`, `gofmt -l` clean, `go test ./...` **281 packages ok / 0 fail**, `make verify-boundary` passes — all re-run *after* merging `origin/main`, not before. Regenerating the boundary manifest produces no diff, which matters because the incoming commits deleted a file under `core/pkg/kernel`, inside the protected surface.

Beyond the suite, this branch was put through an independent review pass and every guard was mutation-proven. Highlights that are worth trusting more than a green tick:

- **25 mutations** were applied to production code and to the analyzers themselves; every one but two turned its intended guard red, and those two were then fixed.
- The cardinality fix was verified by a second pass that stood up a **real in-process OTLP/gRPC receiver** and read the exported protobuf directly: 50 distinct `Host` values, exactly one datapoint per instrument.
- Streaming survives the new layer, checked over a real listener: `Flusher`, `Hijacker`, `CloseNotifier` and `io.ReaderFrom` all preserved — the receipt-tail SSE route depends on it.
- 1,280 requests across 32 concurrent clients plus 400 in-process goroutines: no cross-talk, no race, `go test -race` clean on all three packages.
- Two guards that **could not fail** were caught and fixed. The `core/pkg/api` conversions are now pinned by call *correctness* rather than call shape, so `WriteInternalR(w, nil, err)` is caught — it kept the count green and panicked in production. And the mutation proof subtracts an unmutated baseline, so a walker blinded to the whole tree no longer reports PASS.

## What this does NOT deliver

**The programme's headline acceptance is not met by this branch.** A log record and a span from the same request still do not join on `TraceId`, because the kernel emits logfmt and the id lands as a text field rather than in the column. That needs [HELM-475](https://linear.app/mindburn/issue/HELM-475). Both changes are required; neither is sufficient.

The live acceptance on `dev-3` — build the image, deploy into the slot, drive a request, read `telemetry.otel_traces` back — has not been run. It is Sergey's and ready to go the same day there is an image.

Recorded rather than fixed, all cheap follow-ups: a panicking handler yields a span with no route, status or duration datapoint; one early-return path in `runServerWithOptions` skips the drain entirely; and the chart's `serviceName` pattern uses ASCII `\S` where the Go consumer uses `strings.TrimSpace`, so a leading NBSP validates and is then silently trimmed.

The downstream OSS sync (`make sync-oss-kernel`, `[OSS-SYNC]`) is deliberately **not** attempted here.

## Where to look first

1. `core/cmd/helm-ai-kernel/main.go` — `buildAPIHandler`, the chain and why the position matters.
2. `core/pkg/tracing/edgehandler.go` — the three surfaces, one normalizer, and the catch-all rule.
3. `core/pkg/observability/observability.go` — the metric view and the flush budget split.
4. `core/cmd/helm-ai-kernel/context_logging_contract_test.go` — the `go/ast` contract over three packages, and the four ways a conversion can look done and carry nothing.

Human review is specifically wanted on the middleware ordering and on anything touching route registration.

## Ownership

DRI for [HELM-495](https://linear.app/mindburn/issue/HELM-495) is now Andrei — the diagnosis was the infra-shaped part and it is finished; what remains is kernel Go. Sergey keeps the live acceptance on the stand. Full context in `my-docs/TASKS/HELM-453-observability-foundation/REVIEW-FIXES-2026-08-10.md`.


mindburnlabs (changes requested):
Reviewed exact head d0418b30d5f71464c5953eb1720d60380b91292d.

Blocking:

1. The public HTTP edge still adopts any inbound `traceparent` as the parent, and the tests explicitly require that behavior (`edgehandler.go`, `serve_path_tracing_test.go`). This lets an unauthenticated caller choose the trace ID in kernel logs and clear the sampled bit to suppress the server span. HELM-515 is marked Done in Linear, but it has no recorded decision/comment and its own acceptance requires the opposite on the untrusted edge, with trusted parenting preserved only for mesh `ext_authz`. Please implement an explicit default-deny/public-endpoint trust policy, preserve the trusted mesh path separately, and remove or justify the stale `ext_authz` re-extraction.

2. Required CI is red. `tracked-file-hygiene` rejects the literal TODO tokens added in `context_logging_contract_test.go`. `quantum-crypto-inventory` also requires `quantum_posture` annotations for `core/pkg/api/trust_keys_handler.go` and `core/pkg/credentials/handlers.go`.

3. Non-blocking but worth closing here: the metric-view tests prove the attribute allowlist but not the load-bearing claim that nil `Stream.Aggregation` preserves the recommended duration histogram buckets. One manual-reader assertion on bucket boundaries would prevent a silent future regression.

What looks sound: route normalization/root catch-all handling, the Host-header cardinality allowlist, middleware placement relative to RequestID/rate limiting, request-context conversions, and shutdown budget/error propagation. Focused Go tests pass. Live QA/dev acceptance is still required after a corrected, green exact head; this branch should not go to stage.

mindburnlabs (approved):
Reviewed corrected exact head da523a39114c1c7a9887b0bf4728771a814f19a2. The public daemon edge now starts a new root span, links (rather than adopts) inbound trace context, ignores remote sampling for local span creation, and drops untrusted baggage. The shared-listener ext_authz path no longer re-extracts traceparent; the regression test proves Guardian.EvaluateDecision is a child of the kernel server span. The literal placeholder hygiene failures, quantum_posture inventory gaps, and duration-histogram bucket proof are also closed. Local evidence: affected Go packages pass; make verify-presentation, make verify-boundary (1107 entries), and quantum inventory pass. Approval covers source correctness; exact-head CI and live QA/store correlation remain required before merge/deployment.

mindburnlabs (approved):
Reviewed exact head b4a29598c354f728aca1af9ca53c1d01b73d14b2 against base 2cd3e3cc0b70bc1922ea3d85b62913c628a96d29. The tracing/security implementation is unchanged from the independently reviewed corrected head; the only added delta is the already-reviewed #834 workflow pin. Public/shared-listener behavior is fail-closed root+link, ext_authz no longer bypasses the server span, and Guardian is its child. Local full quality-pr and focused telemetry tests passed. Approval covers source correctness; exact-head CI must finish, and QA/store correlation remains mandatory before HELM-495 or the release is declared complete.

PR #840, opened 2026-08-10
fix(release): make published drift retry executable

mindburnlabs:
## Root cause

The v0.8.3 release published its artifacts, but post-release-version-drift failed because the workflow passed the published-only --surface-timeout option before the published argparse subcommand. Every retry exited during argument parsing instead of checking public surfaces.

## Fix

- place published before --surface-timeout
- strengthen the release workflow contract test to lock the executable argument order

## Evidence

- reproduced from release run 31330205611 / job 93295444919
- focused release workflow contract: 7/7 pass
- corrected CLI form reaches the published checks rather than argparse failure
- diff check clean

## Release gate

This must merge before creating the Kernel v0.8.4 tag. It does not publish or deploy anything by itself.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1396,8 +1396,8 @@ jobs:
           for attempt in $(seq 1 40); do
             if python3 scripts/release/check_version_drift.py \
               --expected-version "$VERSION" \
-              --surface-timeout 10 \
-              published; then
+              published \
+              --surface-timeout 10; then
               exit 0
             fi
             if [ "$attempt" -lt 40 ]; then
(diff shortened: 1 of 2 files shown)


mindburnlabs:
Exact-head checkpoint after guarded update onto #839: head 708318e7f743357546103673adcf2b9b6d2ad46b, base f7c97d827a363f11206dd59b32c988fa6f1e928d. Local impact-aware quality-pr is fully green, including release-workflow-contract, version-drift, docs truth, tracked hygiene, quantum inventory, and contract-breaking self-test. The corrected command reaches published checks; a live 0.8.4 docs-only probe now fails on the three real docs surfaces rather than argparse, which is being remediated in a separate docs deployment lane. Kernel v0.8.4 tag remains on hold until this PR and the docs gate land.

PR #870, opened 2026-08-13
fix(metrics): make kernel RED metrics reachable and separate OTLP metric push

Hirama:
Closes the mechanical half of HELM-477. Both defects in the issue are fixed by the same change, which is why they were filed together.

## 1. The RED families were unreachable

The kernel had two disjoint metric surfaces and served only one. `/metrics` now serves `prometheus.DefaultGatherer` and the MeterProvider's prometheus reader **alongside** the hand-written governance families, so one scrape of the endpoint the chart already points at carries:

| family | why it was missing |
| -- | -- |
| `http_server_request_duration_seconds` | recorded by otelhttp into a MeterProvider that only existed when `OTEL_EXPORTER_OTLP_ENDPOINT` was set, and then only exported over OTLP — a transport HELM-469 had already ruled out |
| `helm_kernel_decisions_total` | a live `promauto` counter incremented on every PDP decision; `/metrics` was the hand-written writer and never gathered the default registry |
| `go_*`, `process_*` | the kernel had no runtime families at all — no heap, goroutines, GC or open FDs |

**On the four RED instruments the issue names** (`helm.requests.total`, `helm.errors.total`, `helm.request.duration`, `helm.operations.active`): they are *not* what this exposes. They have no non-test caller and never had one — `RecordRequest`/`RecordError`/`RecordDuration` are called only from `observability_test.go`, and `InstrumentOperation` has zero callers — so publishing them would have produced permanently-zero families. otelhttp's `http.server.*` instruments are recorded on every request and carry the same semantics, so they are what the pull surface serves.

Live output from the daemon-level test:

```
helm_kernel_decisions_total{backend="test",reason_code="OK",verdict="ALLOW"} 1
http_server_request_duration_seconds_count{http_request_method="GET",http_response_status_code="204",http_route="/api/v1/probe/{id}"} 1
http_server_request_duration_seconds_count{http_request_method="GET",http_response_status_code="500",http_route="/api/v1/boom"} 1
```

Rate, errors and duration, broken down by route and status, with `server_address`/`server_port` absent — the HELM-495 cardinality bound survives the new reader, and there is a test asserting it does.

## 2. An OTLP endpoint for traces silently switched metrics on

Metric export is now selected by the standard `OTEL_METRICS_EXPORTER`, **defaulting to `none`** rather than the spec's `otlp`, because HELM-469 settled the kernel's metric transport as pull-only — defaulting to `otlp` would restore exactly the failure this issue records. The pull surface is unconditional either way: `none` silences the push, not `/metrics`.

The chart gains `telemetry.metricsExporter` (schema-validated `none|otlp`), so the telemetry block is finally enablable for the kernel without dragging a metric push along — which is what was blocking the trace half of HELM-453.

## Ordering, which is load-bearing

Unlike the tracer, which otelhttp resolves lazily per request, the **meter is bound once inside `otelhttp.NewHandler`** (`config.go` `newConfig` → `semconv.NewHTTPServer(c.Meter)`). A handler built before the global MeterProvider exists keeps the no-op meter for the life of the process. `NewServices` already runs before `buildAPIHandler`, so the daemon was already correct; the constraint is now documented where someone can trip over it.

The OTel reader registers into a registry owned by `Services`, not `prometheus.DefaultRegisterer`, so constructing `Services` twice in one process cannot fail on a duplicate collector registration.

## Evidence

- `go test ./pkg/...` — 275 packages, all pass
- `make test-cli`, `-race` on the three touched packages — pass
- `docs-truth`, `verify-boundary`, `go vet`, `gofmt`, `go mod tidy` — clean
- `helm template` with the value on/off/absent, and the schema rejecting a bogus value

Both central tests are mutation-proven:

- disable the pull reader → `no http_server_request_duration family on the pull surface; got []`
- restore the endpoint-implies-push coupling → `OTLP metric exports = 1, want 0`

`make quality-pr` reports two failures locally, both exit 127 from missing tooling on this machine (`ripgrep` for `verify-presentation`, bash 4 `mapfile` for `go_mod_tidy_check.sh`). The underlying rules were checked by hand and pass: no forbidden patterns in the changed files, and `go mod tidy` is idempotent.

## Left to the DRI, as the issue reserves

The orphaned **guardian** and **effects** instrument groups. The issue says wire or delete; wiring is not a switch-flip. `guardian.go:230` declares the `otel` field and never reads it — Guardian calls the global tracer directly at `:868` and `:1403` — so `WithOTel()` currently enables nothing and there are zero internal call sites to turn on. Making them real means writing new instrumentation into the decision hot path, and the decision counting it would add already exists as `helm_kernel_decisions_total`, which this PR makes reachable. Deleting them removes exported API (`WithOTel`, `WithOTelInstrumentation`, `NewEffectsOTelInstrumentation`) that downstream repos may consume, so it wants an owner's call rather than mine. My recommendation is delete.

The `go_*`/`process_*` question the issue asks to record a decision on is answered by this PR: they appear.


mindburnlabs:
Superseded by release replacement #890. Accepted replacement commits aedfd11cd7f85a190d93ad4c1a7d49f2b69e63c8 and cbabbcd57eb55eba85e0ef60a4e37ba86ab8a8d1 are contained by the reviewed v0.8.5 candidate; this branch and history are preserved.

PR #878, opened 2026-08-18
evidence(HELM-618): first live savings EvidencePack — negative parity result (Inference.net)

peycheff-com:
## What

The first **live, API-billed** Savings EvidencePack produced by the machinery merged in #877, plus its capture artifacts and a claim-window fix to the unattributed-spend guard.

- `reference_packs/spend-savings/savingspack-h618-infnet-20260818/` — the pack (manifest `sha256:261cc3cf…f457`, signed by `spend-proxy-d11333ad0bbd`).
- `reference_packs/spend-savings/capture-h618-infnet-20260818/` — task-split manifest (with the frozen selection), task results, parity pre-run.
- `examples/spend-proxy/dogfood.inference-net.config.json` — price book from the official catalog page, captured 2026-08-18.
- `docs/guides/savings-evidencepack-walkthrough.md` — placeholders replaced with the real pack's values; every command block was executed against this pack.

## The result: parity bar FAILED, and that is the deliverable

| route | model | holdout | CPST (token-priced) |
|---|---|---|---|
| baseline | `gpt-4.1` | 16/16 passed | 6675 micro-cents/task |
| substitute | `gpt-4.1-nano` | 12/16 passed | 465 micro-cents/task |

The substitute costs ~14x less per successful task but does not hold quality parity on the holdout, so **no savings claim is made** and `savings_claim_valid` is `false`. Per the [05] methodology this negative result is recorded honestly rather than suppressed — a meter that cannot say "no" is not a meter.

Worth noting: on the 8-task selection set the substitute *tied* the baseline (6/8 each) and was the cheapest candidate meeting the bar, which is why the freeze picked it; the 16-task holdout then refuted it. That is exactly the confound the frozen holdout exists to catch.

## Capture facts

- 65 governed calls total (1 pre-freeze smoke, 32 parity across 4 routes, 32 holdout across 2 routes).
- **$0.0081** actual token-priced spend; 65 cents settled against the $20 ceiling (the ledger ceils sub-cent calls — both bases are in the view).
- Parity pre-run committed (`c518b0285`) and the selection freeze committed (`15f79d01b`) **before** any holdout call; the verifier enforces `frozen_at` < every holdout quote.

## Guard fix

The unattributed-spend guard refused the export because of the pre-capture smoke call. It is now **claim-window aware**: settled spend inside the window (post-freeze) still fails closed, while spend that provably predates the freeze — which cannot enter CPST — is counted and disclosed (`pre_window_unattributed_calls`, plus a stderr note). Regression tests cover both directions.

## Verification

`spend-proxy savings-verify` passes all eleven checks on the committed pack, and the binary-independent walkthrough (shasum/jq/python re-derivations) passes end to end.

## Note on DigitalOcean

DO Gradient was the original target and its config/price book stay in-repo for a future capture; DO serverless inference is prepaid-only and the workspace is on monthly billing, so it is blocked on an account arrangement, not on code.

## Testing

`make test`, `make lint`, `verify-boundary`, quantum inventory, focused + race tests — all green.

Linear: HELM-618


mindburnlabs:
Superseded by release replacement #890. Evidence/source replacement f711bac9ff2ba5ae3c8caa76fd9b9405819f2c5b is retained and bounded by b503a4f9bd168b9a4d5fa94f5e4dfda5b8101de0 in the reviewed v0.8.5 candidate; this branch and history are preserved.

PR #903, opened 2026-08-29
HELM-658: keep unchanged boundary snapshots recovery-stable

mindburnlabs:
## Summary

- avoid rewriting `boundary_surface_snapshots.updated_at` when the canonical snapshot JSON is unchanged
- preserve normal snapshot updates for real boundary state changes
- prove an unchanged SQL registry reload leaves the persisted timestamp stable

## End-to-end finding

The exact seven-component OrganizationRuntime started healthy, but its recovery proof failed closed with `recovery invariant changed: postgres`. The only logical row changed by the restart was the Kernel boundary snapshot timestamp; the snapshot payload was unchanged.

## Validation

- focused unchanged-reload SQLite regression: passed
- full `pkg/boundary` suite: passed
- real PostgreSQL boundary persistence test: passed
- full Kernel Core `go test ./...`: passed

Local/no-effect runtime proof only; no deployment or production promotion is included.

### DIFF
--- a/core/pkg/boundary/surface_registry.go
+++ b/core/pkg/boundary/surface_registry.go
@@ -1443,7 +1443,8 @@ func (r *SurfaceRegistry) persistLocked() error {
 		}
 		_, err := r.db.ExecContext(ctx, `INSERT INTO boundary_surface_snapshots (id, snapshot_json, updated_at)
 			VALUES ($1, $2, $3)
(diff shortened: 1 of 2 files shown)


mindburnlabs:
Superseded by #902, which merged the same recovery-stability fix first and includes exact seven-component recovery plus offline receipt verification. Closing this duplicate rather than resolving it into redundant logic.

